PR #1, opened 2015-05-26
Use py.test ini file integration instead of rolling our own

davehunt:
* Also changes `privacy` tag to be clearly related to Sauce Labs job visibility.
* Also migrates the Sauce Labs and BrowserStack documentation.

### DIFF
--- a/pytest_selenium/cloud/saucelabs.py
+++ b/pytest_selenium/cloud/saucelabs.py
@@ -2,10 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-from ConfigParser import ConfigParser
-import os
 import json
 
+from _pytest.mark import MarkInfo
 from py.xml import html
 import pytest
 import requests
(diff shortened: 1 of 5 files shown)


davehunt:
Pinging @bobsilverberg for review.

pytest_selenium/cloud/saucelabs.py
```diff
@@ -77,12 +62,13 @@ def start_driver(item, options, capabilities):
...
         capabilities['version'] = options.browser_version
-    executor = 'http://%s:%s@ondemand.saucelabs.com:80/wd/hub' % _credentials()
+    executor = 'http://%s:%s@ondemand.saucelabs.com:80/wd/hub' % \
+        _credentials(item.config)
     return webdriver.Remote(command_executor=executor,
                             desired_capabilities=capabilities)
 
 
-def url(session):
+def url(config, session):
```
bobsilverberg: `config` isn't used in this `url` method.

pytest_selenium/pytest_selenium.py
```diff
@@ -110,11 +110,6 @@ def pytest_runtest_makereport(__multicall__, item, call):
     extra_summary = []
     report = __multicall__.execute()
     extra = getattr(report, 'extra', [])
-    try:
-        report.public = item.keywords['privacy'].args[0] == 'public'
```
bobsilverberg: I see `report.public` being removed, but not replaced with anything. Was it not actually being used for anything before?

pytest_selenium/cloud/saucelabs.py
davehunt: No, but it's passed to it.

Dave Hunt

pytest_selenium/pytest_selenium.py
```diff
@@ -136,29 +131,50 @@ def pytest_runtest_makereport(__multicall__, item, call):
...
 
 
 def pytest_addoption(parser):
-    config = ConfigParser.ConfigParser(defaults={'baseurl': ''})
-    config.read('mozwebqa.cfg')
+    parser.addini('selenium_base_url', 'base url for selenium')
+
+    # browserstack configuration
+    parser.addini('browserstack_username',
+                  'browserstack username',
```
bobsilverberg: Below (for sauce labs config) you include the `help=` argument name, which makes it clearer what the second argument is, but here you leave it out. I think for consistency and clarity you should add it here as well.

pytest_selenium/pytest_selenium.py
davehunt: Yeah, it was an artefact from the log publicity checking of RC.

Dave Hunt

pytest_selenium/pytest_selenium.py
davehunt: +1 thanks.

Dave Hunt

bobsilverberg:
Very nice @davehunt! Great to be able to use code built into py.test instead of writing our own to parse the config file.

r+ wc

davehunt:
Merged in https://github.com/davehunt/pytest-selenium/commit/6cdd173fdde851b5c9171370fa99851e52731682

Issue #5, opened 2015-06-11
Clean up command line options


davehunt:
I'd like to revisit and clean up the existing command line options. Here are the current options, with some suggestions for improvements:

* **--baseurl** - Rename to --base-url
* **--skipurlcheck** - I think we should remove this. I believe it was added for testing sites that are not http(s) such as about:support, etc. I would be inclined to reimplement this as a protocol check in _verify_base_url
* **--host** - leave this as it is
* **--port** - leave this as it is
* **--driver** - leave this as it is for now, but I'd like to have a local default - we could potentially determine which browsers are present and conveniently pick one. I don't think we should assume the user will have a Selenium server running
* **--capability** - leave this as it is
* **--chromepath** - rename to --driver-path
* **--firefoxpath** - rename to --driver-path
* **--firefoxpref** - rename to --firefox-pref
* **--profilepath** rename to --firefox-profile
* **--extension** rename to --browser-extension
* **--chromeopts** - rename to repeatable --chrome-option argument
* **--operapath** - rename to --driver-path
* **--browsername** - remove and document using --capability for Grid, Sauce Labs, and BrowserStack
* **--browserver** - remove and document using --capability for Grid, Sauce Labs, and BrowserStack
* **--platform** - remove and document using --capability for Grid, Sauce Labs, and BrowserStack
* **--build** - replace with environment variable (BUILD_ID in Jenkins, TRAVIS_BUILD_NUMBER in Travis, etc)
* **--untrusted** - remove and revisit in the future if it's needed
* **--proxyhost** - rename to --proxy-host
* **--proxyport** - rename to --proxy-port
* **--eventlistener** - rename to --event-listener
* **--sensitiveurl** - rename to --sensitive-url and allow configuration from environment variable
* **--destructive** - leave this as it is

davehunt:
I'm keen for feedback here. /cc @bobsilverberg, @justinpotts, @codingjoe

codingjoe:
I think it would be great to offer most options as an environment variable (security) or pytest-conf variable as well.

davehunt:
All command line options *should* be able to be set using PYTEST_ADDOPTS environment variable, however I agree that it's nice to have dedicated variable names for most. What is the pytest-conf you mention, is that a plugin (I couldn't find it).

davehunt:
[...]

PR #11, opened 2015-06-12
Updated command line options

justinpotts:
Issue #5


pytest_selenium/pytest_selenium.py
```diff
@@ -207,59 +200,32 @@ def pytest_addoption(parser):
...
-                     help='target browser version (default: %default).')
-    group._addoption('--platform',
-                     action='store',
-                     default=os.environ.get('SELENIUM_PLATFORM'),
+    group._addoption('--chrome-option',
+                     action='append',
+                     dest='chrome_option',
                      metavar='str',
-                     help='target platform (default: %default).')
+                     help='google chrome option to set.')
```
davehunt: add `in format "name:value"` to the help text.

davehunt:
Tests are failing due to the changes to the command line options. For example https://github.com/justinpotts/pytest-selenium/blob/comline_options/testing/conftest.py#L21 needs to be updated to `--base-url`.

pytest_selenium/pytest_selenium.py
```diff
@@ -207,59 +200,32 @@ def pytest_addoption(parser):
                      metavar='str',
                      help='additional capability to set in format '
                           '"name:value".')
-    group._addoption('--chromepath',
+    group._addoption('--driver-path',
                      action='store',
-                     dest='chrome_path',
+                     dest='driver_path',
```
davehunt: We'll need to update the following to refer to the new option name:
https://github.com/justinpotts/pytest-selenium/blob/comline_options/pytest_selenium/driver.py#L69
https://github.com/justinpotts/pytest-selenium/blob/comline_options/pytest_selenium/driver.py#L79
https://github.com/justinpotts/pytest-selenium/blob/comline_options/pytest_selenium/driver.py#L97

pytest_selenium/pytest_selenium.py
davehunt: We'll also need to modify https://github.com/justinpotts/pytest-selenium/blob/comline_options/pytest_selenium/driver.py#L146 to work with the new format. Looking at that, we may also need a `--chrome_argument` or even `--driver-argument` command line option depending on how generic we can be.

davehunt: I'd be okay with you dropping this out of the scope of this issue and raising a separate issue to consider how we should handle driver options and arguments. Ideally we'd avoid too many browser specific options.

davehunt:
@justinpotts I've just landed https://github.com/davehunt/pytest-selenium/commit/36f8bcb9b69e6c9cc90113c8adf2564d167f5d17 which will cause your branch to need rebasing and likely some conflicts. Let me know if you need any help resolving these.

davehunt:
Thanks @justinpotts this looks great. There are a couple of issues running against a remote server, so I'll pick this up now and will request review of my changes from you shortly.

PR #17, opened 2015-06-24
Fixes #12 - Remove the default driver

### LINKED ISSUE #12: Remove the default driver
Currently we default to the 'Remote' driver, which could suggest a preference of starting a Selenium server. Instead I would suggest removing the default, and failing with a list of accepted drivers if none is specified.

bobsilverberg:
I'm not sure if we wanted to try to introspect the `driver.py` file to determine the valid values of `--driver`, but I figured not. I also realize that the new test I added doesn't really belong in `test_remote.py` but I didn't see anywhere else to put it. I suppose we could create a new test file for tests that are not specific to a driver.

@davehunt / @justinpotts r?

### DIFF
--- a/pytest_selenium/driver.py
+++ b/pytest_selenium/driver.py
@@ -11,6 +11,15 @@ from selenium.webdriver.common.proxy import Proxy
 from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 from selenium import webdriver
 
+SUPPORTED_DRIVERS = ['BrowserStack',
+                     'Chrome',
+                     'Firefox',
+                     'IE',
+                     'Opera',
+                     'PhantomJS',
+                     'Remote',
+                     'SauceLabs']
+
 
 def proxy_from_options(options):
     if options.proxy_host and options.proxy_port:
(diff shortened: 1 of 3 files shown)


bobsilverberg:
@davehunt @justinpotts I added a new commit that addresses what we discussed. r?

pytest_selenium/driver.py
```diff
@@ -11,6 +11,9 @@
 from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 from selenium import webdriver
 
+SUPPORTED_DRIVERS = ['BrowserStack', 'Chrome', 'Firefox', 'Ie',
+                     'Opera', 'PhantomJS', 'Remote', 'SauceLabs']
```
davehunt: Personal preference to put each item on a new line. I'm also tempted to rename `Ie` as `IE` because it looks nasty.

bobsilverberg: I think it looks `nice`.

pytest_selenium/driver.py
```diff
@@ -44,9 +47,13 @@ def make_driver(item):
 
 def start_driver(item, capabilities):
     options = item.config.option
-    specific_driver = '%s_driver' % options.driver.lower()
-    driver_method = globals().get(specific_driver, generic_driver)
-    driver = driver_method(item, capabilities)
+    driver_name = (options.driver and options.driver.lower()
+                   or None)
+    if driver_name not in [browser.lower() for browser in SUPPORTED_DRIVERS]:
```
davehunt: I'd prefer `if options.driver and options.driver.lower() not in [d for d in SUPPORTED_DRIVERS]`

davehunt: Actually, give me a moment...

bobsilverberg: Seriously, `browser`? I must be tired.

bobsilverberg: The problem with not creating a variable that stores the value of `options.driver and options.driver.lower()` is that on a later line we need to use `options.driver` again, so we'd need to do that check for `None` again. I'd rather do it once, which is why I created the variable.

davehunt: `options.driver.lower() in map(str.lower, SUGGESTED_DRIVERS)`

PR #22, opened 2015-06-26
Support specifying capabilities by file

davehunt:


pytest_selenium/pytest_selenium.py
```diff
@@ -189,17 +197,10 @@ def pytest_addoption(parser):
                           '(default: %default)')
     group._addoption('--driver',
                      action='store',
+                     choices=SUPPORTED_DRIVERS,
```
bobsilverberg: This will mean the case of the driver specified on the command line must exactly match the case of the choices defined. Are we okay with that? The previous code was more flexible about case.

pytest_selenium/pytest_selenium.py
```diff
@@ -189,17 +197,10 @@ def pytest_addoption(parser):
...
                      dest='driver',
-                     metavar='str',
-                     help='webdriver implementation. '
-                          'Valid values are: %s.' %
-                          ', '.join(SUPPORTED_DRIVERS))
-    group._addoption('--capability',
```
bobsilverberg: I thought we were going to keep the `--capability` command line argument to be used for simple strings.

setup.py
```diff
@@ -10,6 +10,7 @@
       packages=['pytest_selenium', 'pytest_selenium.cloud'],
       install_requires=[
           'pytest>=2.2.4',
+          'pytest-variables',
```
bobsilverberg: Did you already release a version of this that supports multiple files?

pytest_selenium/pytest_selenium.py
davehunt: Yeah, I was a bit concerned at first, but I think I like it being strict. :)

pytest_selenium/pytest_selenium.py
davehunt: I explained that I have one more commit for this PR.. the behaviour of this will change but it will be restored. :)

testing/test_capabilities.py
```diff
@@ -0,0 +1,34 @@
...
+        import pytest
+        @pytest.fixture
+        def capabilities():
+            return {'foo': 'bar'}
+    """)
+    file_test = testdir.makepyfile("""
+        import pytest
+        @pytest.mark.nondestructive
+        def test_capabilities(capabilities):
+            assert True
```
bobsilverberg: Why `assert True`? Why not check `capabilities`?

setup.py
davehunt: No, but there's a PR open now.

testing/test_capabilities.py
davehunt: Fixed. :)

davehunt:
Ready for review now.

bobsilverberg:
LGTM @davehunt. I particularly like the fact that now (after some other PRs were merged) all of the tests pass for me. I no longer have to manually ignore the Opera and Chrome failures! :)

r+

davehunt:
Thanks @bobsilverberg, this has now been merged.

PR #27, opened 2015-08-19
Reintroduce support for specifying the path to the driver executable

davehunt:
I couldn't think of a smarter way to allow users to specify the paths to driver executables. Pinging @bobsilverberg for review.

### DIFF
--- a/pytest_selenium/driver.py
+++ b/pytest_selenium/driver.py
@@ -5,10 +5,11 @@
 import copy
 
 import pytest
+from selenium import webdriver
+from selenium.webdriver.common.desired_capabilities import DesiredCapabilities
+from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
 from selenium.webdriver.support.event_firing_webdriver import \
     EventFiringWebDriver
-from selenium.webdriver.firefox.firefox_binary import FirefoxBinary
-from selenium import webdriver
 
 
 def start_driver(item, capabilities):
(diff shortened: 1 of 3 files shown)


davehunt:
I've merged this in https://github.com/davehunt/pytest-selenium/commit/d75656319317fc70768ccae499c8ea1c7dbc70ed but I'd still appreciate a post-merge review if you have any feedback or suggestions.

pytest_selenium/driver.py
```diff
@@ -38,31 +39,43 @@ def browserstack_driver(item, capabilities):
...
+        kwargs['executable_path'] = options.driver_path
+    return webdriver.Chrome(**kwargs)
 
 
 def firefox_driver(item, capabilities):
     """Return a WebDriver using a Firefox instance"""
     options = item.config.option
-    binary = None
-    if options.firefox_path:
+    kwargs = {'capabilities': capabilities or None}
```
bobsilverberg: Why `capabilities or None` here, but just `capabilities` at https://github.com/davehunt/pytest-selenium/pull/27/files#diff-c37349b36e3bdd7a136af78712541351R43

pytest_selenium/driver.py
```diff
@@ -38,31 +39,43 @@ def browserstack_driver(item, capabilities):
...
+        kwargs['firefox_binary'] = FirefoxBinary(options.firefox_path)
+    kwargs['firefox_profile'] = _create_firefox_profile(options)
+    return webdriver.Firefox(**kwargs)
 
 
 def ie_driver(item, capabilities):
     """Return a WebDriver using an Internet Explorer instance"""
-    return webdriver.Ie(capabilities=capabilities)
+    options = item.config.option
+    kwargs = {'capabilities': capabilities or None}
```
bobsilverberg: Same question here.

pytest_selenium/driver.py
```diff
@@ -38,31 +39,43 @@ def browserstack_driver(item, capabilities):
...
+    if options.driver_path is not None:
+        kwargs['executable_path'] = options.driver_path
+    return webdriver.Ie(**kwargs)
 
 
 def phantomjs_driver(item, capabilities):
     """Return a WebDriver using a PhantomJS instance"""
-    return webdriver.PhantomJS(desired_capabilities=capabilities)
+    options = item.config.option
+    _capabilities = capabilities or DesiredCapabilities.PHANTOMJS
```
bobsilverberg: Where did `DesiredCapabilities.PHANTOMJS` come from? Is that something that was missing from the previous version but should have been there? Could you address this by making the default value of the `capabilities` argument `DesiredCapabilities.PHANTOMJS`, rather than having to use this `_capabilities` intermediary variable?

bobsilverberg:
Looks good for the most part @davehunt. I added a couple of questions to the diff.

pytest_selenium/driver.py
davehunt: Selenium explicitly handles `None`, the bug here is that `chrome_driver` should probably include `or None`.

pytest_selenium/driver.py
davehunt: Unfortunately the PhantomJS driver does not handle `None` for capabilities. This is likely a bug in Selenium as it's inconsistent with the other driver implementations.

PR #28, opened 2015-08-19
Remove verify autouse fixture

davehunt:
@birdsarah would you mind taking a look at this? It moves the verifying of the base URL into the default base URL fixture. This means that by overriding the base_url fixture, there will be a responsibility on the fixture author to reimplement the verification if they choose.

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -41,26 +41,24 @@ def _environment(request, base_url, capabilities):
 
 @pytest.fixture(scope='session')
 def base_url(request):
-    """Return base URL from command line options"""
+    """Return a verified base URL"""
     config = request.config
-    url = config.option.base_url or config.getini('selenium_base_url')
-    if not url:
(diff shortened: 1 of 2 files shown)


davehunt:
Also, thinking about your comments on disabling the safety checks, I think with this patch it becomes fairly straightforward, but would be a good idea to document.

  1. override the `base_url` fixture to disable the URL verification
  1. `--destructive` command line, `addopts`, or `PYTEST_ADDOPTS`
  1. `--sensitive-url=None` command line, `addopts`, `PYTEST_ADDOPTS`, or `SELENIUM_SENSITIVE_URL`

birdsarah:
This is a definite improvement - thanks - haven't run it locally, but am assuming it works.

Regarding docs, I have no idea what `sensitive-url` is, and can't see it in the docs - can you elaborate?

davehunt:
Merged in https://github.com/davehunt/pytest-selenium/commit/209a68d92e371fbe779e0873ec1012adb4639297. I'll add details of the sensitive URL checking to the docs today - thanks for the prompt! :)

PR #29, opened 2015-08-19
Avoid checking if the base URL is senstive for every test

davehunt:
I realised we were checking the history of the reponses for every test, which is excessive. This addresses that and also cleans up the fixture names. If you want to disable sensitive URL checking you can either set the `--sensitive-url` option to `None` or override the `sensitive_url` fixture and return `False`.

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -73,7 +73,7 @@ def capabilities(request, variables):
 
 
 @pytest.fixture
-def selenium(request, _sensitive_skipping, capabilities):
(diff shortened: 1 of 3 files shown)


davehunt:
@birdsarah mind taking a look over this one too? :)

birdsarah:
I like the rename to `sensitive_url` - seems more clear to me how it all fits together. Don't have any other opinions as have never used this feature.

davehunt:
Thanks, merged in https://github.com/davehunt/pytest-selenium/commit/70400c69be5c4c49ad20b0e4403b2ce625edd69f. I have a couple more tweaks in this area that I'll work on today.

PR #30, opened 2015-08-20
Add a way to control when HTML report debug is gathered, and capture logs unless excluded

davehunt:
Also added a section to the documentation on the HTML report.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -356,3 +356,26 @@ the implicit timeout or window size, you can override the ``selenium`` fixture:
       selenium.implicitly_wait(10)
       selenium.maximize_window()
       return selenium
+
+HTML Report
+***********
+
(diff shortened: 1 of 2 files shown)


davehunt:
@bobsilverberg would you mind reviewing this?

davehunt:
I've merged this in https://github.com/davehunt/pytest-selenium/commit/768d0e41c520f60f4ee15c032b67ba015aca87b2 but would still appreciate a post-merge review if you have any feedback or suggestions.

bobsilverberg:
@davehunt I just looked through the diff and it looks good to me. Nice additions. I notice there aren't any tests added for these changes. I assume you tested it all manually and were satisfied with that.

davehunt:
Originally I didn't have tests because `pytest-html` wasn't a dependency. Now it is, I think it makes sense to add tests for the customisation, but I have yet to write these tests.

PR #35, opened 2015-10-08
Strip trailing slashes from base URL

davehunt:
I'd appreciate your thoughts on this change @bobsilverberg. My thinking was that we could strip trailing forward slashes from the base URL so that there's a consistent expectation of how to construct URLs from it. I'm conflicted though as we're manipulating something explicitly set by the user, which could be confusing and in some circumstances frustrating. I think I've talked myself out of landing this, but would like to hear your thoughts.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -1,6 +1,10 @@
 Release Notes
 =============
 
+**1.0b3**
+
+* Strip trailing slashes from base URL.
+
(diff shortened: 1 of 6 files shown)


bobsilverberg:
r-, based on our discussion in IRC

davehunt:
Thanks. :smile:

PR #42, opened 2015-11-27
Consistently stash the base URL in the configuration options

davehunt:
@bobsilverberg could you review this? It will mean that we only need to check `config.getoption` for the base URL even if it's specified in an INI file.

See also: https://github.com/mozilla/mcom-tests/blob/e0618d117000f338c37ebabd96de52469a4ca39b/tests/test_l10n_download_links.py#L13 and https://github.com/mozilla/Socorro-Tests/pull/341/files#r45722474

/cc @m8ttyB

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -118,6 +118,23 @@ responding before proceeding with the test suite. To enable this, specify the
 ``--verify-base-url`` command line option or set the ``VERIFY_BASE_URL``
 environment variable to ``TRUE``.
 
+Skipping Base URLs
+------------------
+
+You can `skip tests <http://pytest.org/latest/skipping.html>`_ based on the
+value of the base URL so long as it is provided either by the command line or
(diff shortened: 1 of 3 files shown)


bobsilverberg:
This looks good, @davehunt. r+

Would this be worth adding to the docs as a sort of "pro tip" - to let folks know where to look for the `base_url` if they need it?

davehunt:
In the process of adding the documentation I discovered this wasn't quite doing what we needed. I've now fixed it, and added a new test that verifies we can skip base on a base URL provided by a configuration file. @bobsilverberg r?

bobsilverberg:
Looks good, but Travis is failing.

davehunt:
I've seen a number of failures in Travis for BrowserStack connections timing out. I think we can safely ignore them, but I've now rebased this so we'll see if we get a pass this time.

PR #43, opened 2015-11-27
Report warnings when gathering debug fails rather than raising exceptions

davehunt:
Also I don't know why, but we never previously attempted to gather debug on setup/teardown, which could be useful (especially on setup when fixtures are involved). Would you mind having a look over this @bobsilverberg?

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -7,7 +7,6 @@ import os
 
 import pytest
 import requests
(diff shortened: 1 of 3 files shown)


bobsilverberg:
Very nice @davehunt. This is indeed much better!  r+

davehunt:
Merged in https://github.com/davehunt/pytest-selenium/commit/dd122106471da48ceaa209edb0bf56514fc0f3cf

PR #47, opened 2015-12-17
Set a timeout for all requests calls and warn if any cloud provider calls fail

davehunt:
@bobsilverberg mind having a look over these changes? I noticed that if our tests fail due to the remote session ending we can timeout waiting to update the test status. This just makes sure we always set a reasonable timeout for requests, and also makes sure that any issues communicating with a cloud provider do not cause exceptions, but are logged as warnings.

### DIFF
--- a/pytest_selenium/cloud/browserstack.py
+++ b/pytest_selenium/cloud/browserstack.py
@@ -40,7 +40,8 @@ class Provider(CloudProvider):
     def url(self, config, session):
         response = requests.get(self._session_url.format(id=session),
                                 auth=(self._username(config),
-                                      self._access_key(config)))
+                                      self._access_key(config)),
+                                timeout=10)
         return response.json()['automation_session']['browser_url']
 
     def additional_html(self, session):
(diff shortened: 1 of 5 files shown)


bobsilverberg:
@davehunt This looks pretty good. I have a couple of questions:

1. Why did you decide to go with a hardcoded `10` in multiple locations rather than keep the `REQUESTS_TIMEOUT` that you had in `pytest_selenium.py` and then import from there?

2. Did you check whether there is a way to set a default timeout for any call to `requests`? If there is setting that once might save all of these `timeout=` arguments.

Other than that it lgtm. I haven't run any tests. Let me know if you'd like me to.

davehunt:
> Why did you decide to go with a hardcoded `10` in multiple locations rather than keep the `REQUESTS_TIMEOUT` that you had in `pytest_selenium.py` and then import from there?

Because I couldn't get the imports to work for me. I'll revisit this, as I'd prefer not to hard-code the timeouts.

> Did you check whether there is a way to set a default timeout for any call to `requests`? If there is setting that once might save all of these `timeout=` arguments.

I've checked a number of times and they still haven't implemented this. As far as I can tell it's a bug that the global socket timeout is not used, which would be better than never timing out. See https://github.com/kennethreitz/requests/issues/2228 for some background.

davehunt:
I've decided to merge this as is for now. I do think we can improve this in the future either by having the timeout set centrally, or by allowing the user to configure the timeout. The best scenario would be for requests to either support setting a default timeout, or for it to use the global socket timeout.

PR #49, opened 2016-01-19
Only update the status for cloud providers after the test method

davehunt:
Without this, the cloud provider status is updated after setup, test method, and teardown. This can cause the cloud provider to report that the test has passed when in reality the test fails but the teardown is successful. I'm thinking there may be a better solution to this that would allow failures in setup and teardown to be reported to the cloud providers, but I think that would require a fairly major refactor.

Pinging @bobsilverberg for review.


bobsilverberg:
LGTM @davehunt. I tested it locally and saw a failure posted as a success to SauceLabs without the patch, and then properly as a failure with the patch.

r+, and thanks for the investigation and the fix!

davehunt:
This has been superseded by #51

PR #64, opened 2016-04-12
Update according to change log

icemac:
It says that Python 2.6 support was dropped in version 1.1.


davehunt:
Support for __pytest__ 2.6 was dropped, but not __Python__ 2.6. :smile:

icemac:
Sorry for the noise.

PR #66, opened 2016-06-21
removed session scope on 3 methods (pytest_selenium._verify_url, safet…

Maxyme:
I found compatibility issues when using pytest-selenium with pytest-tornado.
Running the pytest-tornado with pytest-selenium installed would crash:
https://github.com/eugeniy/pytest-tornado
With the scope removed, everything should be working again.

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -57,7 +57,7 @@ def base_url(request):
         return base_url
 
 
-@pytest.fixture(scope='session', autouse=True)
+@pytest.fixture
 def _verify_url(request, base_url):
     """Verifies the base URL"""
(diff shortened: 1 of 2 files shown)


Maxyme:
I've been digging into it a bit more. It seems that the problem lies with the base_url fixture that both pytest-tornado and pytest-selenium use. Perhaps this PR isn't exactly the way to go.

davehunt:
The scope is necessary, although I have moved the base URL implementation to [pytest-base-url](https://pypi.python.org/pypi/pytest-base-url/). I still need to update pytest-selenium to depend on the new plugin. Perhaps this will resolve your issue too. Another option would be to reimplement these fixtures with different scopes in your project.

PR #69, opened 2016-07-07
Add fixture to override firefox path

justinpotts:


pytest_selenium/drivers/firefox.py
```diff
@@ -47,6 +46,11 @@ def firefox_driver(request, capabilities, driver_path, firefox_profile):
 
 
 @pytest.fixture
+def firefox_path(request):
+    return request.config.getoption('firefox-path')
```
davehunt: This should be `firefox_path` rather than `firefox-path`, though I suspect both might work.

davehunt:
Looks good. Could you just add a note to the release notes in pytest-selenium/blob/master/docs/news.rst

davehunt:
Merged in https://github.com/pytest-dev/pytest-selenium/commit/7eb6fa067d773d3963ff9b17319be72519e43851 thanks @justinpotts!

PR #88, opened 2016-11-02
fix issue: AttributeError: 'module' object has no attribute 'ie'

MungoWang:
To fix error: AttributeError: 'module' object has no attribute 'ie' when using --driver IE

### DIFF
--- a/pytest_selenium/drivers/__init__.py
+++ b/pytest_selenium/drivers/__init__.py
@@ -1,1 +1,1 @@
-from . import internet_explorer as ie
+from . import internet_explorer as ie  # noqa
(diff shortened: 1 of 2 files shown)


davehunt:
I've been waiting for this! Someone using pytest-selenium to test on Windows with Internet Explorer or Edge! :)

pytest_selenium/drivers/__init__.py
```diff
@@ -0,0 +1 @@
+from . import internet_explorer as ie
```
davehunt: Can you fix this so flake8 doesn't report an unused warning?

MungoWang: Sorry that, I didn't realize the problem, fixed now.

davehunt:
Thanks! I've merged, and tagged v1.5.1 so there should be a new release available soon.

Issue #59, opened 2016-03-22
Session scoped selenium fixture


germanp:
Hi!

Is there any way to use only one selenium instance for running various tests? I found that the selenium fixture is scoped by default as function level. But I could not found any way to overtake this using any pytest functionality.

Thanks

ProProgrammer:
Instantiate the selenium driver in conftest.py as a fixture and set scope to `session` to use same driver instance throughout the session.
That's how I use it at my work.

ProProgrammer:
This is not an issue. Please close it if this solves your query.

davehunt:
This is also how I would do it too, although I always recommend a fresh browser instance for each test. If you reuse the browser you risk session state affecting later tests.

germanp:
Sorry, but I couldn't catch the idea. What do you mean by instantiate the selenium driver? Do you mean pasting the implementations from python_selenium.drivers? 

I've tried several ways and I always get ScopeMismatch exception. The only way I found to work was to change the scope in the plugin's code.

Thanks!

ProProgrammer:
Hi,
I meant having a fixture like this in your `conftest.py` file.
```
@pytest.fixture(scope="session")
def driver(request):
    _driver = webdriver.Firefox()
    _driver.implicitly_wait(200)
    _driver.maximize_window()
    _driver.get('https://www.google.com')
    request.addfinalizer(_driver.quit)
    return _driver
```

davehunt:
Here's an alternative `conftest.py` solution that doesn't require you to instantiate your own WebDriver instance:

```python
import pytest


@pytest.fixture(scope='session')
def selenium(request):
    from pytest_selenium import selenium
    return selenium(request)


@pytest.fixture(scope='session')
def firefox_driver(request, capabilities, driver_path, firefox_profile):
    from pytest_selenium.drivers.firefox import firefox_driver
    return firefox_driver(request, capabilities, driver_path, firefox_profile)


@pytest.fixture(scope='session')
def capabilities(request, session_capabilities):
    from pytest_selenium import capabilities
    return capabilities(request, session_capabilities)


@pytest.fixture(scope='session')
def driver_path(request):
    from pytest_selenium import driver_path
    return driver_path(request)


@pytest.fixture(scope='session')
def firefox_profile(request):
    from pytest_selenium.drivers.firefox import firefox_profile
    return firefox_profile(request)
```

Note that I've just used the Firefox driver here, if you needed another driver you'd have to override more fixtures.

[...]

PR #91, opened 2016-11-28
[WIP] Module-scoped fixtures

feedingaliencat:
Work in progress: checking Travis results

The aim of this PR is to permit keeping the browser open between test-cases, instead of open and close it for each test case.
The proposed solution is to make the selenium fixture module-scoped instead of function-scoped.

### DIFF
--- a/pytest_selenium/drivers/firefox.py
+++ b/pytest_selenium/drivers/firefox.py
@@ -49,7 +49,7 @@ def firefox_path(request):
     return request.config.getoption('firefox_path')
 
 
-@pytest.fixture
+@pytest.fixture(scope='session')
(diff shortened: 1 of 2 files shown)


davehunt:
Thank you for the pull request, however unless you find a way to implement this without changing the default behaviour this will not be merged. See https://github.com/pytest-dev/pytest-selenium/issues/59 for more details on the reasons why and an alternate approach you could use. I'll leave this open for now as you've indicated that it's a work in progress.

feedingaliencat:
Thank you @davehunt for pointing me to the solution proposed here: https://github.com/pytest-dev/pytest-selenium/issues/59#issuecomment-200338319

PR #93, opened 2017-01-24
Move sensitive configuration options into separate files

### LINKED ISSUE #60: Move sensitive configuration options into separate files
At the moment usernames and keys for services like Sauce Labs, BrowserStack and TestingBot can either be set via environment variables or in the shared pytest configuration file. Using the configuration file means both private and public options are shared. It can also be difficult to manage when the configuration file is checked into source control and you don't want to expose your credentials to others.

I think it makes sense to add support for new configuration files such as `.saucelabs`, `.browserstack`, and `.testingbot` to hold these sensitive values. These should be able to be stored in the project directory (and can easily be ignored from version control), any parent directory, or the user's home directory. The files should be discovered in this order and the first match should take priority, allowing projects to override user defaults.

davehunt:
Read credentials for cloud testing services from configuration files in the working or home directory. Fixes #60

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -1,6 +1,32 @@
 Release Notes
 =============
 
+**1.8.0 (unreleased)**
+
+* **BREAKING CHANGE:** Moved cloud testing provider credentials into separate
+  files for improved security.
+
+  * If you are using the environment variables for specifying cloud testing
+    provider credentials, then you will not be affected.
+  * If you are storing credentials from any of the cloud testing providers in
+    one of the default configuration files then they will no longer be used.
+    These files are often checked into source code repositories, so it was
+    previously very easy to accidentally expose your credentials.
+  * Each cloud provider now has their own configuration file, such as
+    ``.browserstack``, ``.crossbrowsertesting``, ``.saucelabs``,
+    ``.testingbot`` and these can be located in the working directory or in the
+    user's home directory. This provides a convenient way to set up these files
+    globally, and override them for individual projects.
+  * To migrate, check ``pytest.ini``, ``tox.ini``, and ``setup.cfg`` for any
+    keys starting with ``browserstack_``, ``crossbrowsertesting_``,
+    ``saucelabs_``, or ``testingbot_``. If you find any, create a new
+    configuration file for the appropriate cloud testing provider with your
+    credentials, and remove the entries from the original file.
+  * The configuration keys can differ between cloud testing providers, so
+    please check the :doc:`user_guide` for details.
+  * See `#60 <https://github.com/pytest-dev/pytest-selenium/issues/60>`_ for
+    for original issue and related patch.
+
 **1.7.0 (2016-11-29)**
 
 * Introduced a ``firefox_options`` fixture.
--- a/pytest_selenium/drivers/browserstack.py
+++ b/pytest_selenium/drivers/browserstack.py
@@ -2,46 +2,56 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-import os
-
 import pytest
 import requests
 
-DRIVER = 'BrowserStack'
-API_JOB_URL = 'https://www.browserstack.com/automate/sessions/{session}.json'
-EXECUTOR_URL = 'http://{username}:{key}@hub.browserstack.com:80/wd/hub'
+from pytest_selenium.drivers.cloud import Provider
 
 
-def pytest_addoption(parser):
-    parser.addini('browserstack_username',
-                  help='browserstack username',
-                  default=os.getenv('BROWSERSTACK_USERNAME'))
-    parser.addini('browserstack_access_key',
-                  help='browserstack access key',
-                  default=os.getenv('BROWSERSTACK_ACCESS_KEY'))
+class BrowserStack(Provider):
+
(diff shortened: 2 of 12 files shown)


pytest_selenium/drivers/browserstack.py
```diff
@@ -2,6 +2,12 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+try:
+    import configparser
+except ImportError:
```
nicoddemus: Hey @davehunt a small comment here: in my experience it is better to try to import using a version check rather than fallback from an import error, because the latter might make it harder to debug some import errors. For example, if one has a botched installation or someone has mocked `configparser` somehow and it fails to import, it will fallback to the second import and "hide" the original error (in Python 2 at least, in Python 3 it won't be a big deal). I had this issue in pytest-dev/pytest-mock#68, where `import mock` was failing because some missing dependencies.

In summary, I suggest using this instead:

```python
if sys.version_info[0] == 2:
    import ConfigParser as configparser
else:
    import configparser
```

pytest_selenium/drivers/browserstack.py
```diff
@@ -63,24 +60,26 @@ def pytest_selenium_runtest_makereport(item, report, summary, extra):
...
-    username = config.getini('browserstack_username')
+def credentials():
+    config = configparser.ConfigParser()
+    config.read(['.browserstack', os.path.expanduser('~/.browserstack')])
+    try:
+        username = config.get('credentials', 'username')
+    except (configparser.NoSectionError, configparser.NoOptionError, KeyError):
+        username = os.getenv('BROWSERSTACK_USERNAME')
     if not username:
         raise pytest.UsageError('BrowserStack username must be set')
```
nicoddemus: It would be nice to mention `BROWSERSTACK_USERNAME` in this error message.

pytest_selenium/drivers/browserstack.py
```diff
@@ -63,24 +60,26 @@ def pytest_selenium_runtest_makereport(item, report, summary, extra):
...
-    return username
+    try:
+        key = config.get('credentials', 'key')
+    except (configparser.NoSectionError, configparser.NoOptionError, KeyError):
+        key = os.getenv('BROWSERSTACK_ACCESS_KEY')
+    if not key:
+        raise pytest.UsageError('BrowserStack access key must be set')
```
nicoddemus: Ditto on this error message as well.

pytest_selenium/drivers/saucelabs.py
```diff
@@ -71,27 +67,29 @@ def driver_kwargs(request, test, capabilities, **kwargs):
...
-def _api_key(config):
-    api_key = config.getini('sauce_labs_api_key')
-    if not api_key:
-        raise pytest.UsageError('Sauce Labs API key must be set')
-    return api_key
-
-
-def _username(config):
-    username = config.getini('sauce_labs_username')
+def credentials():
```
nicoddemus: Perhaps it would be worth refactoring `credentials` so it can be reused by the various drivers? The logic seems to be identical across them: read a config file, try to get the credentials from it, fallback to an environment variable, etc.

testing/test_crossbrowsertesting.py
```diff
@@ -33,12 +34,37 @@ def failure(testdir, testfile, httpserver_base_url):
...
+
+def test_missing_access_key_file(failure, monkeypatch, tmpdir):
+    credentials = tmpdir.join('.crossbrowsertesting')
+    credentials.write('[credentials]\nusername=foo')
+    monkeypatch.setattr(os.path, 'expanduser', lambda p: str(credentials))
+    assert 'UsageError: CrossBrowserTesting auth key must be set' in failure()
+
+
+def test_invalid_credentials_env(failure, monkeypatch, tmpdir):
+    monkeypatch.setattr(os.path, 'expanduser', lambda p: str(tmpdir))
```
nicoddemus: If you refactor `credentials()` so it can be reused, I suggest then refactoring the tests as well so you can use the same set of tests related to credentials in all the drivers. 

I usually do this by having a base class with test methods, and then subclass that base class in each module configuring it (perhaps using class variables) with the specifics for that driver.

davehunt:
@nicoddemus thank you so much for the review... your suggestions were spot on! I've addressed the duplication in the cloud drivers, but have left the test duplication for now. I've also introduced a much more helpful custom exception. This refactoring will allow for further improvements, but I'll leave those for another time. If you get a chance, I'd appreciate another review.

docs/news.rst
```diff
@@ -1,6 +1,14 @@
 Release Notes
 =============
 
+**1.8.0 (unreleased)**
+
+* **BREAKING CHANGE:** Moved cloud testing provider credentials into separate
+  files for improved security.
```
nicoddemus: I think it would be nice to mention the changes in a little more detail, perhaps mentioning #60, and a little example of what users should do to migrate their current configuration.

davehunt: Thanks, I've made some updates and rebased/squashed in preparation for a merge and release later today. I really appreciate your time @nicoddemus!

nicoddemus: Absolutely, my pleasure! 😁 

Thanks for all the time and effort you put in your excellent plugins as well! 👍

PR #100, opened 2017-03-13
Update selenium to session level

jlane9:
Having selenium setup and tear down each time a test is ran is
expensive. Set selenium driver instead to be applied at the session-level.


davehunt:
Hi @jlane9, thanks for the pull reuqest, however this is not going to be merged. Having setup/teardown per test is expensive but intentional, and not doing so can cause tests to fail in unpredictable ways due to long running browser instances and unknown leftover state from previous tests. This has been discussed before (#59), and if you really want a per session browser instance I suggest overwriting the fixtures in a local `conftest.py` or even packaging your own plugin that extends this one.

Issue #103, opened 2017-03-17
Add support for edge


birdsarah:

davehunt:
This should be pretty easy. As an example, Safari support was added in 75092cb6b0891e76453b2ebf2f9094254aea21a2 with a test in 03cd43e5e647b9e1c03e1d7da191cb6964537e14 and driver path support was added in https://github.com/pytest-dev/pytest-selenium/commit/52c78ac4cb221487d4897e3aef70ce901143be34.

birdsarah:
Am working on it.

PR #104, opened 2017-03-18
Add edge support

birdsarah:
Fixes: #103

Have tested this on a windows machine and works fine.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -142,6 +142,23 @@ option to indicate where it can be found:
 See the `ChromeDriver documentation <https://sites.google.com/a/chromium.org/chromedriver/>`_
 for more information.
 
+
+Edge
+----
+
+To use Edge, you will need to
+`download Edge WebDriver <https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/tools/webdriver>`_
+and specify ``Edge`` for the ``--driver`` command line option. If the driver
+executable is not available on your path, you can use the ``--driver-path``
+option to indicate where it can be found:
+
+.. code-block:: bash
+
+  $ py.test --driver Edge --driver-path /path/to/edgedriver
+
+Selenium v2.47 or greater is required.
+
+
 Internet Explorer
 -----------------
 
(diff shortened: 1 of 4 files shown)


birdsarah:
Not sure what these test failures are - would appreciate some guidance.

davehunt:
@birdsarah these failures look to be related to the Selenium python client issue that you've addressed: `AttributeError: 'str' object has no attribute 'decode'`. I think we can ignore them for now, and hopefully get a new client release out soon.

davehunt:
Could you add a test for Edge? We won't be able to run it on Travis, but might ultimately be able to get it to run on AppVeyor. We'll at least be able to run it locally when on a Windows machine with Edge and EdgeDriver. See the Safari test for an example: https://github.com/pytest-dev/pytest-selenium/commit/03cd43e5e647b9e1c03e1d7da191cb6964537e14

birdsarah:
(just so you know I'll be away for 10 days - will get back to this early april)

docs/user_guide.rst
```diff
@@ -142,6 +142,23 @@ option to indicate where it can be found:
...
+
+To use Edge, you will need to
+`download Edge WebDriver <https://docs.microsoft.com/en-us/microsoft-edge/dev-guide/tools/webdriver>`_
+and specify ``Edge`` for the ``--driver`` command line option. If the driver
+executable is not available on your path, you can use the ``--driver-path``
+option to indicate where it can be found:
+
+.. code-block:: bash
+
+  $ py.test --driver Edge --driver-path /path/to/edgedriver
```
davehunt: I've recently removed the `bash` from the command line example in the documentation. Also, the path should probably be `\path\to\MicrosoftWebDriver.exe` to represent the actual executable name and a Windows path.

docs/user_guide.rst
```diff
@@ -142,6 +142,23 @@ option to indicate where it can be found:
...
+
+Selenium v2.47 or greater is required.
```
davehunt: pytest-selenium depends on Selenium 3.x already, so we can remove this note.

birdsarah:
Hi, sorry for delay. This week's looking busy.....trying to be realistic....feel free to jump in.

davehunt:
Closing this in favour of #121, which addresses my comments. It took me a while to get a VM up and running but I was able to test this successfully. Unfortunately there's no way to test Edge in CI at the moment, so the test will have to be run manually if there's a risk of regression.

Issue #108, opened 2017-04-05
Add support for geckodriver log


jrbenny35:
I would like to get the geckodriver log when there is a geckodriver related failure, and have it added to the html report.

I can work on this but would need a little mentoring.

davehunt:
Thanks for raising this @jrbenny35, I agree it would be a valuable addition. By default the geckodriver log file is named geckodriver.log and written to the current working directory, which is useful for anyone running locally wanting to check the contents. I think there are a couple of options:

1. Attempt to locate geckodriver.log when enhancing the HTML report. If we find it, we can add it to the report. The problem with this is that it makes some assumptions on the log file name and location is specific to Firefox/geckodriver.

2. We could explicitly tell Selenium where to write the log file, and store this in the pytest configuration object. Then, when we enhance the HTML report we just look for this attribute in the configuration object and if found we can add the log.

I'd say let's go for option 2. First, we'd want to add `log_path` as a driver keyword argument in [this fixture](https://github.com/pytest-dev/pytest-selenium/blob/7d7008256a8a53d5dc3f237d130af774ed43771f/pytest_selenium/drivers/firefox.py#L34-L41). I don't think we need to create a `driver_log` fixture, but that's also an option. For the value we should use 'geckodriver.log' in the working directory, so that there's no unexpected changes from the current behaviour. We should then be able to store this as `pytestconfig._driver_log` (you'll need to add the `pytestconfig` fixture to the argument list). Then, when [gathering logs](https://github.com/pytest-dev/pytest-selenium/blob/7d7008256a8a53d5dc3f237d130af774ed43771f/pytest_selenium/pytest_selenium.py#L200-L217) for the HTML report, we should append the file located at `item.config._driver_log` if it's present. This approach means we should easily be able to add support for other drivers that allow users to specify a log location.

Let me know if you have any questions.

jrbenny35:
The function ```pytest_runtest_makereport``` is what generates the HTML?

davehunt:
Nope, the [pytest-html](https://pypi.python.org/pypi/pytest-html) plugin generates the HTML. The `pytest_runtest_makereport` hook just adds stuff to `report.extra` to be picked up by the other plugin.

jrbenny35:
Trying to run a test of this command and I keep running into this error

```selenium.common.exceptions.WebDriverException: Message: GET /session/729f9e4e-b687-2445-91bb-1368f159748e/log/types did not match a known command```

[...]

PR #109, opened 2017-04-10
Initial attempt at adding geckodriver logs

jrbenny35:
@davehunt 
I have made an attempt at solving Issue #108.

The logs will appear in a link tab named ```Geckodriver log```. If the log is not available it is printed out. I think this could get annoying however.

I also moved the variable that sets up ```pytest-html``` to the top of the ```_gather_logs``` function since I use that in the geckodriver log append as well.

### DIFF
--- a/pytest_selenium/drivers/firefox.py
+++ b/pytest_selenium/drivers/firefox.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import warnings
+import os
 
 import pytest
 from selenium.webdriver import FirefoxProfile
@@ -33,12 +34,15 @@ def pytest_addoption(parser):
                      help='path to a firefox extension.')
 
 
-def driver_kwargs(capabilities, driver_path, firefox_options, **kwargs):
+def driver_kwargs(capabilities, driver_path, firefox_options, log_path,
+                  pytestconfig, **kwargs):
     kwargs = {}
     if capabilities:
         kwargs['capabilities'] = capabilities
     if driver_path is not None:
         kwargs['executable_path'] = driver_path
+    log_path = os.path.realpath('geckodriver.log')
+    kwargs['log_path'] = pytestconfig._driver_log = log_path
     kwargs['firefox_options'] = firefox_options
     return kwargs
 
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -59,7 +59,8 @@ def driver_args():
 
 @pytest.fixture
 def driver_kwargs(request, capabilities, chrome_options, driver_args,
-                  driver_class, driver_path, firefox_options, firefox_profile):
+                  driver_class, driver_path, firefox_options, firefox_profile,
+                  pytestconfig):
     kwargs = {}
     driver = request.config.getoption('driver').lower()
     kwargs.update(getattr(drivers, driver).driver_kwargs(
@@ -72,6 +73,8 @@ def driver_kwargs(request, capabilities, chrome_options, driver_args,
         host=request.config.getoption('host'),
         port=request.config.getoption('port'),
         request=request,
+        log_path=None,
+        pytestconfig=pytestconfig,
         test='.'.join(split_class_and_test_names(request.node.nodeid))))
     return kwargs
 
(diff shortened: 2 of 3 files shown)


pytest_selenium/drivers/firefox.py
```diff
@@ -31,12 +31,14 @@ def pytest_addoption(parser):
...
 
-def driver_kwargs(capabilities, driver_path, firefox_options, **kwargs):
+def driver_kwargs(capabilities, driver_path, firefox_options, log_path,
+                  pytestconfig, **kwargs):
     kwargs = {}
     if capabilities:
         kwargs['capabilities'] = capabilities
     if driver_path is not None:
         kwargs['executable_path'] = driver_path
+    pytestconfig._driver_log = log_path
```
davehunt: Where is `log_path` implemented? We also need to pass the path to the log as a keyword argument. This means adding something like:

```python
log_path = os.path.realpath('geckodriver.log')
kwargs['log_path'] = pytestconfig._driver_log = log_path
```

pytest_selenium/pytest_selenium.py
```diff
@@ -64,6 +64,8 @@ def driver_kwargs(request, capabilities, driver_class, driver_path,
         host=request.config.getoption('host'),
         port=request.config.getoption('port'),
         request=request,
+        log_path=os.path.realpath('geckodriver.log'),
```
davehunt: Ahh.. so this is where `log_path` is being set.. This is Firefox specific so it shouldn't be set here. We *could* have a `driver_log` fixture, but I don't think we need one at this point.

pytest_selenium/pytest_selenium.py
```diff
@@ -198,6 +200,15 @@ def _gather_html(item, report, driver, summary, extra):
 
 
 def _gather_logs(item, report, driver, summary, extra):
+    pytest_html = item.config.pluginmanager.getplugin('html')
+    # Open and read geckodriver log
+    if item.config._driver_log is not None:
+        with open(item.config._driver_log, 'r') as driver_log:
+            extra.append(pytest_html.extras.text(
+                driver_log.read(), 'Geckodriver Log'))
```
davehunt: This shouldn't be implementation specific, as other drivers may also provide logs. Let's call it 'Driver Log'.

pytest_selenium/pytest_selenium.py
```diff
@@ -198,6 +200,15 @@ def _gather_html(item, report, driver, summary, extra):
...
+    else:
+        summary.append('WARNING: Failed to gather gecodriver log')
```
davehunt: There's no need for an else here.

jrbenny35:
@davehunt could you take another look?

pytest_selenium/pytest_selenium.py
```diff
@@ -198,6 +200,13 @@ def _gather_html(item, report, driver, summary, extra):
 
 
 def _gather_logs(item, report, driver, summary, extra):
+    pytest_html = item.config.pluginmanager.getplugin('html')
+    # Get geckodriver log if available
+    if item.config._driver_log is not None:
```
davehunt: I would expect this to raise an exception if the attribute is not set (it's only set for Firefox driver). We should probably use `hasattr` or `getattr` instead.

pytest_selenium/pytest_selenium.py
```diff
@@ -198,6 +200,13 @@ def _gather_html(item, report, driver, summary, extra):
 
 
 def _gather_logs(item, report, driver, summary, extra):
+    pytest_html = item.config.pluginmanager.getplugin('html')
+    # Get geckodriver log if available
```
davehunt: Remove this comment, as it's specific to the Firefox driver.

pytest_selenium/drivers/firefox.py
```diff
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+import os
```
davehunt: Nit: Please keep the blank line before the first import.

jrbenny35:
Should the test go into ```test_report.py``` ?

davehunt:
> Should the test go into `test_report.py`?

Works for me. Alternatively, create `test_report_driver_log.py`.

davehunt:
Thanks @jrbenny35! I merged this in https://github.com/pytest-dev/pytest-selenium/commit/f54fef89f42509bfe8a9cee4100b1e81a80590d7 and followed up with an enhancement that adds support to additional drivers.

Issue #96, opened 2017-02-13
Check additional environment variables for credentials


davehunt:
When using Jenkins declarative pipelines, credentials can be set as environment variables as follows:

```groovy
environment { 
    SAUCELABS = credentials('SAUCELABS') 
}
```

For credentials which are of type "Standard username and password", the environment variable specified will be set to `username:password` and two additional environment variables will be automatically be defined: `MYVARNAME_USR` and `MYVARNAME_PSW` respectively.

It would be convenient for pytest-selenium to therefore check `SAUCELABS_USR` and `SAUCELABS_PSW` for Sauce Labs, and similar for `BROWSERSTACK`, `TESTINGBOT`, and `CROSSBROWSERTESTING`.

See https://jenkins.io/doc/pipeline/tour/environment/ for more details.

BeyondEvil:
I can probably do this at the same time as #107

BeyondEvil:
ping @davehunt

davehunt:
@BeyondEvil I've invited you to collaborate on the project so I can assign this to you. Please feel free to start work on this, however I'd prefer a separate pull request.

PR #110, opened 2017-04-22
Add jenkins declarative pipeline env variables

BeyondEvil:
I did my best to update the user guide/documentation as well.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -222,16 +222,19 @@ a ``.saucelabs`` configuration file in the working directory or your home
 directory, or by setting the ``SAUCELABS_USERNAME`` and ``SAUCELABS_API_KEY``
 environment variables.
 
-Alternatively, when using `Jenkins CI <https://jenkins.io/>`_ declarative piplines,
+Alternatively, when using `Jenkins CI`_ declarative pipelines,
 credentials can be set as environment variables as follows:
 
-.. code-block:: ini
+.. code-block:: groovy
 
   environment {
     SAUCELABS = credentials('SAUCELABS')
   }
 
-For more information see Jenkins `docs <https://jenkins.io/doc/pipeline/tour/environment/>`_.
+For more information see Jenkins `docs`_.
+
+.. _Jenkins CI: <https://jenkins.io/>
+.. _docs: <https://jenkins.io/doc/pipeline/tour/environment/>
 
 Configuration
 ~~~~~~~~~~~~~
@@ -269,16 +272,16 @@ a ``.browserstack`` configuration file in the working directory or your home
 directory, or by setting the ``BROWSERSTACK_USERNAME`` and
 ``BROWSERSTACK_ACCESS_KEY`` environment variables.
 
-Alternatively, when using `Jenkins CI <https://jenkins.io/>`_ declarative piplines,
+Alternatively, when using `Jenkins CI`_ declarative pipelines,
 credentials can be set as environment variables as follows:
 
-.. code-block:: ini
+.. code-block:: groovy
 
   environment {
     BROWSERSTACK = credentials('BROWSERSTACK')
   }
 
-For more information see Jenkins `docs <https://jenkins.io/doc/pipeline/tour/environment/>`_.
+For more information see Jenkins `docs`_.
 
 Configuration
 ~~~~~~~~~~~~~
@@ -314,18 +317,18 @@ a ``.testingbot`` configuration file in the working directory or your home
 directory, or by setting the ``TESTINGBOT_KEY`` and ``TESTINGBOT_SECRET``
 environment variables.
 
-Alternatively, when using `Jenkins CI <https://jenkins.io/>`_ declarative piplines,
+Alternatively, when using `Jenkins CI`_ declarative pipelines,
 credentials can be set as environment variables as follows:
 
-.. code-block:: ini
+.. code-block:: groovy
 
   environment {
     TESTINGBOT = credentials('TESTINGBOT')
   }
 
-Note that for TestingBot, ``username`` corresponds to ``secret`` and ``password`` to ``key``.
+Note that for TestingBot, ``username`` corresponds to ``key`` and ``password`` to ``secret``.
 
-For more information see Jenkins `docs <https://jenkins.io/doc/pipeline/tour/environment/>`_.
+For more information see Jenkins `docs`_.
 
 Configuration
 ~~~~~~~~~~~~~
@@ -363,16 +366,16 @@ a ``.crossbrowsertesting`` configuration file in the working directory or your
 home directory, or by setting the ``CROSSBROWSERTESTING_USERNAME`` and
 ``CROSSBROWSERTESTING_AUTH_KEY`` environment variables.
 
-Alternatively, when using `Jenkins CI <https://jenkins.io/>`_ declarative piplines,
+Alternatively, when using `Jenkins CI`_ declarative pipelines,
 credentials can be set as environment variables as follows:
 
-.. code-block:: ini
+.. code-block:: groovy
 
   environment {
     CROSSBROWSERTESTING = credentials('CROSSBROWSERTESTING')
   }
 
-For more information see Jenkins `docs <https://jenkins.io/doc/pipeline/tour/environment/>`_.
+For more information see Jenkins `docs`_.
 
 Configuration
 ~~~~~~~~~~~~~
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -222,6 +222,17 @@ a ``.saucelabs`` configuration file in the working directory or your home
 directory, or by setting the ``SAUCELABS_USERNAME`` and ``SAUCELABS_API_KEY``
 environment variables.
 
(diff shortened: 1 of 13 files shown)


BeyondEvil:
This is to address issue #96

BeyondEvil:
Re the failure on Travis. I got this failure locally _before_ making any changes. Tried changing the selenium version to 3.3.1 and 3.3.3 but same thing happened. It's only the crossbrowsertesting one that fails, the other providers with the "same" test passes.

I figured it was just something that was weird with my local env. Guess not. :(

davehunt:
> Re the failure on Travis. I got this failure locally *before* making any changes. Tried changing the selenium version to 3.3.1 and 3.3.3 but same thing happened. It's only the crossbrowsertesting one that fails, the other providers with the "same" test passes.

This was due to changes by CrossBrowserTesting. I've updated the tests and triggered a rebuild.

pytest_selenium/drivers/cloud.py
```diff
@@ -38,5 +38,13 @@ def get_credential(self, key, env):
                 KeyError):
             value = os.getenv(env)
```
davehunt: Instead of adding `get_jenkins_credential` we could accept a list for `key`, so we can pass the potential values to look for. For example:

```python
self.get_credential('key', ['TESTINGBOT_USR', 'TESTINGBOT_KEY'])
```

Then we can just loop over these keys and return the first time we find a value. This also means we can set a precedence so `TESTINGBOT_KEY` would override `TESTINGBOT_USR`, which I think makes sense.

pytest_selenium/exceptions.py
```diff
@@ -11,4 +11,5 @@ def __init__(self, driver, key, env):
         super(MissingCloudCredentialError, self).__init__(
             '{0} {1} must be set. Try setting the {2} environment '
             'variable, or see the documentation for how to use a '
-            'configuration file.'.format(driver, key, env))
+            'configuration file. For Jenkins credentials, please '
+            'refer to the documentation'.format(driver, key, env))
```
davehunt: I think we can omit the Jenkins mention here. We're already referring to the documentation, which you've updated to cover Jenkins pipelines.

testing/test_browserstack.py
```diff
@@ -50,10 +50,12 @@ def test_missing_access_key_file(failure, monkeypatch, tmpdir):
     assert 'BrowserStack key must be set' in failure()
 
 
-def test_invalid_credentials_env(failure, monkeypatch, tmpdir):
+@pytest.mark.parametrize(('username', 'key'), [('BROWSERSTACK_USERNAME', 'BROWSERSTACK_ACCESS_KEY'),
+                                               ('BROWSERSTACK_USR','BROWSERSTACK_PSW')])
```
davehunt: Nice use of parametrize! 👍

pytest_selenium/drivers/testingbot.py
```diff
@@ -32,6 +33,12 @@ def key(self):
     def secret(self):
         return self.get_credential('secret', 'TESTINGBOT_SECRET')
 
+    def get_jenkins_credential(self, key):
+        if 'secret' == key:
+            return os.getenv(self.name.upper() + '_USR')
+        elif 'key' == key:
+            return os.getenv(self.name.upper() + '_PSW')
```
davehunt: I believe `secret` should be `_PSW` and `key` should be `_USR`.

BeyondEvil: Yeah, I was thinking that we should be consistent with key... but, your way makes more sense. :)

docs/user_guide.rst
```diff
@@ -222,6 +222,17 @@ a ``.saucelabs`` configuration file in the working directory or your home
...
+Alternatively, when using `Jenkins CI <https://jenkins.io/>`_ declarative piplines,
+credentials can be set as environment variables as follows:
+
+.. code-block:: ini
+
+  environment {
+    SAUCELABS = credentials('SAUCELABS')
+  }
+
+For more information see Jenkins `docs <https://jenkins.io/doc/pipeline/tour/environment/>`_.
```
BeyondEvil: Sure, I do my best. :)

docs/user_guide.rst
```diff
@@ -258,6 +269,17 @@ a ``.browserstack`` configuration file in the working directory or your home
 directory, or by setting the ``BROWSERSTACK_USERNAME`` and
 ``BROWSERSTACK_ACCESS_KEY`` environment variables.
 
+Alternatively, when using `Jenkins CI <https://jenkins.io/>`_ declarative piplines,
```
BeyondEvil: Ah, smaht!

pytest_selenium/drivers/cloud.py
BeyondEvil: It's so obvious now that you suggested it. 😅

BeyondEvil:
@davehunt Why does appveyor fail?

docs/user_guide.rst
```diff
@@ -222,6 +222,20 @@ a ``.saucelabs`` configuration file in the working directory or your home
...
+.. code-block:: groovy
+
+  environment {
+    SAUCELABS = credentials('SAUCELABS')
+  }
+
+For more information see Jenkins `docs`_.
+
+.. _Jenkins CI: <https://jenkins.io/>
+.. _docs: <https://jenkins.io/doc/pipeline/tour/environment/>
```
davehunt: These aren't working for me in GitHub's view. I think it might be the `<>` around the URLs? Also, let's put these at the end of the file, and could you make the link text/reference for the environment docs more descriptive? Something like:

```rst
For more information, see `using environment in Jenkins pipelines`_.

.. _using environment variables in Jenkins pipelines: https://jenkins.io/doc/pipeline/tour/environment/
```

pytest_selenium/drivers/cloud.py
```diff
@@ -30,13 +30,14 @@ def config(self):
...
-                configparser.NoOptionError,
-                KeyError):
-            value = os.getenv(env)
-        if not value:
-            raise MissingCloudCredentialError(self.name, key, env)
-        return value
+    def get_credential(self, key, envs):
+        for env in envs:
+            try:
+                value = self.config.get('credentials', key)
```
davehunt: We don't need to try this for each environment variable, so we should move the for loop into the except clause.

pytest_selenium/drivers/cloud.py
```diff
@@ -30,13 +30,14 @@ def config(self):
...
+            except (configparser.NoSectionError,
+                    configparser.NoOptionError,
+                    KeyError):
+                value = os.getenv(env)
+            if value:
+                return value
+        raise MissingCloudCredentialError(self.name, key, envs)
```
davehunt: It looks like we should reword the message for this custom exception to handle the list of environment variable names.

pytest_selenium/drivers/cloud.py
```diff
@@ -30,13 +30,14 @@ def config(self):
...
-            value = self.config.get('credentials', key)
-        except (configparser.NoSectionError,
-                configparser.NoOptionError,
-                KeyError):
-            value = os.getenv(env)
-        if not value:
-            raise MissingCloudCredentialError(self.name, key, env)
-        return value
+    def get_credential(self, key, envs):
+        for env in envs:
```
davehunt: If `env` is a string, then we'll iterate over each character. Perhaps this isn't an issue as we're always passing a list, and we don't expect users to call this directly.

setup.py
```diff
@@ -50,5 +50,6 @@
           'Programming Language :: Python :: 3.3',
           'Programming Language :: Python :: 3.4',
           'Programming Language :: Python :: 3.5',
+          'Programming Language :: Python :: 3.6',
```
davehunt: It looks like maybe you merged instead of rebased? It's not a problem, I should be able to remove these commits before merging into master.

pytest_selenium/drivers/cloud.py
BeyondEvil: You're the boss. If you want me to handle the case where it is a string let me know how. :)

My .02 are however that if you're going to call this directly you'll also be able to figure out that what you send in needs to be in the form of a list. 👍

pytest_selenium/drivers/cloud.py
BeyondEvil: Yeah, sorry about that. Will fix.

pytest_selenium/drivers/cloud.py
BeyondEvil: Right, I'll give it a shot.

davehunt:
> Why does appveyor fail?

Because I haven't invested the time needed to get it to pass yet. I started it, but hit some issues that might now be resolved.

docs/user_guide.rst
BeyondEvil: dammit, I tried them out in an online rst parser. But maybe the merge/rebase debacle screwed me.

Still not 100% confident with git :(

Sorry.

davehunt:
Merged in https://github.com/pytest-dev/pytest-selenium/commit/aa11135958a875ed0b9aba73eb6f8da42865de33. Thanks @BeyondEvil!

Issue #107, opened 2017-04-03
support for testingbot local tunnel


micheletest:
For local testing, testingbot provides a tunnel which will relay requests to their grid. This means that the driver executor will be localhost:4445/wd/hub instead of hub.testingbot.com:4444/wd/hub.  See documentation here: https://testingbot.com/support/other/tunnel

Currently pytest-selenium doesn't support this. Would you want to? If so, not sure if you have any ideas on how to accomplish it.

davehunt:
The executor URL is currently hard-coded, here: https://github.com/pytest-dev/pytest-selenium/blob/7d7008256a8a53d5dc3f237d130af774ed43771f/pytest_selenium/drivers/testingbot.py#L23-L25

We could add `--host` and `--port` arguments much like we have with the remote driver, here: https://github.com/pytest-dev/pytest-selenium/blob/7d7008256a8a53d5dc3f237d130af774ed43771f/pytest_selenium/drivers/remote.py#L12-L22

These would default to `hub.testingbot.com` and `4444` to maintain the existing functionality, but this will then allow users to configure them for the local tunnel.

BeyondEvil:
@davehunt @micheletest I'm looking at implementing this, but I'm a bit confused regarding the port.

The hardcoded version does not provide a port (implicit 80?) so I'm not sure how to handle that.

https://github.com/pytest-dev/pytest-selenium/blob/7d7008256a8a53d5dc3f237d130af774ed43771f/pytest_selenium/drivers/testingbot.py#L23-L25

BeyondEvil:
ping @davehunt :)

davehunt:
Sorry for the late response. I would suggest making port 80 explicit and the default. I am also against adding new command line arguments for each provider such as `--testingbot-host`. Instead, we can just use `--host`.

davehunt:
This has been fixed in #120 thanks to @BeyondEvil and will be included in the upcoming release (hopefully this week).

PR #111, opened 2017-04-22
Add support for other providers to use non-default host and port

BeyondEvil:
This is a starting point to handle issue #107 

The PR has not been manually tested yet. I wanted you to look at my proposed solution before I went thru the effort of testing it out.

Also, adding automation for this was not straightforward. Need some guidance if I am to accomplish it.

When it comes to documentation/user guide I'm not sure how you want to handle it.

### DIFF
--- a/pytest_selenium/drivers/remote.py
+++ b/pytest_selenium/drivers/remote.py
@@ -3,23 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import os
-
 import pytest
-
-
-def pytest_addoption(parser):
-    group = parser.getgroup('selenium', 'selenium')
-    group._addoption('--host',
-                     default=os.environ.get('SELENIUM_HOST', 'localhost'),
-                     metavar='str',
-                     help='host that selenium server is listening on. '
-                          '(default: %default)')
-    group._addoption('--port',
-                     type='int',
-                     default=os.environ.get('SELENIUM_PORT', 4444),
-                     metavar='num',
-                     help='port that selenium server is listening on. '
-                          '(default: %default)')
 
 
 def driver_kwargs(capabilities, firefox_profile, host, port, **kwargs):
@@ -30,9 +14,18 @@ def driver_kwargs(capabilities, firefox_profile, host, port, **kwargs):
     capabilities.setdefault('version', '')  # default to any version
     capabilities.setdefault('platform', 'ANY')  # default to any platform
 
-    executor = 'http://{host}:{port}/wd/hub'.format(host=host, port=port)
+    executor = 'http://{host}:{port}/wd/hub'.format(host=host or get_host(),
+                                                    port=port or get_port())
     kwargs = {
         'command_executor': executor,
         'desired_capabilities': capabilities,
         'browser_profile': firefox_profile}
     return kwargs
+
+
+def get_host():
+    return os.environ.get('SELENIUM_HOST', 'localhost')
+
+
+def get_port():
+    return os.environ.get('SELENIUM_PORT', 4444)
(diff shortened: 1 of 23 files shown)


pytest_selenium/drivers/remote.py
```diff
@@ -30,9 +14,18 @@ def driver_kwargs(capabilities, firefox_profile, host, port, **kwargs):
     capabilities.setdefault('version', '')  # default to any version
     capabilities.setdefault('platform', 'ANY')  # default to any platform
 
-    executor = 'http://{host}:{port}/wd/hub'.format(host=host, port=port)
+    executor = 'http://{host}:{port}/wd/hub'.format(host=host or get_host(),
```
davehunt: Rather than adding `get_host` and `get_port` let's just do:

```python
host = host or os.environ.get('SELENIUM_HOST', 'localhost')
port = port or os.environ.get('SELENIUM_PORT', 4444)
```

pytest_selenium/drivers/testingbot.py
```diff
@@ -15,14 +15,18 @@ class TestingBot(Provider):
     API = 'https://api.testingbot.com/v1/tests/{session}'
     JOB = 'http://testingbot.com/members/tests/{session}'
 
+    def __init__(self, host=None, port=None):
+        super(TestingBot, self).__init__()
+        self._host = host
+        self._port = port
```
davehunt: Why not just set these directly rather than adding properties:

```python
self.host = host or 'hub.testingbot.com'
self.port = port or 80
```

pytest_selenium/drivers/testingbot.py
```diff
@@ -15,14 +15,18 @@ class TestingBot(Provider):
...
+
     @property
     def auth(self):
         return (self.key, self.secret)
 
     @property
     def executor(self):
-        return 'http://{0}:{1}@hub.testingbot.com/wd/hub'.format(
-            self.key, self.secret)
+        return 'http://{0.key}:{0.secret}@{0.host}:{0.port}/wd/hub'
```
davehunt: I'd prefer to return the formatted executor string rather than the template.

pytest_selenium/pytest_selenium.py
```diff
@@ -265,3 +265,12 @@ def pytest_addoption(parser):
                      metavar='str',
                      help='selenium eventlistener class, e.g. '
                           'package.module.EventListenerClassName.')
+    group._addoption('--host',
+                     metavar='str',
+                     help='host that selenium server is listening on. '
+                          '(default: %default)')
```
davehunt: The defaults are a little more obscured now. We could say something like: "host that the selenium server is listening on, which will default to the cloud provider default or localhost."

davehunt:
> Adding automation for this was not straightforward. Need some guidance if I am to accomplish it.

Do you mean adding tests for the changes? You could add a TestingBot test that sets the host to something invalid and just check that it raises the appropriate exception. I'm not sure what else you could do, within reason.

> When it comes to documentation/user guide I'm not sure how you want to handle it.

Maybe just add a subsection to TestingBot in the user guide for how to configure the host/port when using a local tunnel.

BeyondEvil:
I'm sorry @davehunt. :( I don't know what happened. Tried to fix it, but I clearly have no idea what I'm doing.

BeyondEvil:
Pushed a clean PR in: https://github.com/pytest-dev/pytest-selenium/pull/120

Should I close this one? Don't want the conversation to disappear.

davehunt:
> Tried to fix it, but I clearly have no idea what I'm doing.

There's a reason it's called 'git'! What I usually do is update my master branch from the remote upstream master `git checkout master && git pull --rebase upstream master` and then rebase my feature branch against my local master `git checkout feature && git rebase master`. Then you can push to your origin, though you'll need to force as you're rewriting history. What commands did you use?

davehunt:
> Should I close this one? Don't want the conversation to disappear.

It's not a big change, so I'll just review it fresh.

BeyondEvil:
Bah, I think I tried to pull and rebase `origin featurebranch` against `remote master` or something.

I'll try your way next time, thanks @davehunt !

PR #112, opened 2017-04-23
Add Python 3.6 classifier to setup.py

levi-rs:
Add Python 3.6 classifier to setup.py.

3.6 is currently being used for testing, but the latest supported release per the classifiers is 3.5.


davehunt:
Thanks for the patch @levi-rs! I'm going to fix the Travis CI failures - which are unrelated - before merging this.

PR #118, opened 2017-04-27
remove jenkins function

BeyondEvil:
Finally. @davehunt 

Sorry about that, but at least I fixed it.


davehunt:
Oops! Thanks for spotting that.

PR #120, opened 2017-04-30
Add support for other providers to use non-default host and port [clean]

BeyondEvil:
Pushed a clean one instead. Hope that's OK?

Old one:  #111

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -391,12 +391,10 @@ the ``--capability`` command line arguments. See the
 for full details of what can be configured.
 
 Local tunnel
-____________
+~~~~~~~~~~~~
 
-To run the tests using `TestingBots local tunnel <https://testingbot.com/support/other/tunnel>`_
-you'll also need to set the ``--host`` and ``--port`` command line arguments::
-
-    --host localhost --port 4445
+To run the tests using `TestingBot's local tunnel <https://testingbot.com/support/other/tunnel>`_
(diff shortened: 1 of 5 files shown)


docs/user_guide.rst
```diff
@@ -390,6 +390,14 @@ the ``--capability`` command line arguments. See the
 `test options <http://testingbot.com/support/other/test-options>`_
 for full details of what can be configured.
 
+Local tunnel
+____________
```
davehunt: Let's bump this up a level so it's alongside "Running tests".

docs/user_guide.rst
```diff
@@ -390,6 +390,14 @@ the ``--capability`` command line arguments. See the
...
+
+To run the tests using `TestingBots local tunnel <https://testingbot.com/support/other/tunnel>`_
```
davehunt: Nit: TestingBot's (note the apostrophe). Alternatively 'using a local TestingBot tunnel'.

docs/user_guide.rst
```diff
@@ -390,6 +390,14 @@ the ``--capability`` command line arguments. See the
...
+you'll also need to set the ``--host`` and ``--port`` command line arguments::
+
+    --host localhost --port 4445
```
davehunt: I think we should either give a full command line here, or just skip the example.

BeyondEvil:
@davehunt Done! ^_^

PR #123, opened 2017-05-09
Fix dead link in user guide

BeyondEvil:
The old link is dead and leads to an "does not exist yet" page. The new link redirects to:

https://docs.pytest.org/en/latest/customize.html#command-line-options-and-configuration-file-settings

Should I add that one instead?


davehunt:
Thanks @BeyondEvil!

PR #124, opened 2017-05-18
Use a separate log file for each driver instance

davehunt:


davehunt:
Merged in https://github.com/pytest-dev/pytest-selenium/commit/13c4c997d8d80b7cc8125c11acc5893943de6e13

PR #126, opened 2017-05-30
Doc base url sensitive env

Jenselme:
- Improve documentation about base URL
- Print a message if tests are skipped because of a sensitive URL. Currently, for this to work, the user needs to use the `-s` option. If you have a better solution, please tell me.

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -44,6 +44,12 @@ Specifying a Base URL
 To specify a base URL, refer to the documentation for the
 `pytest-base-url <https://github.com/pytest-dev/pytest-base-url>`_ plugin.
 
+.. note::
+
+  By default, any tests using a base URL will be skipped. This is because all tests
+  are considered destructive, and all environments are considered sensitive. See
+  `Sensitive Environments`_ for further details.
+
 Sensitive Environments
 **********************
 
--- a/pytest_selenium/safety.py
+++ b/pytest_selenium/safety.py
@@ -33,7 +33,13 @@ def pytest_configure(config):
 
 
 def pytest_report_header(config, startdir):
-    return 'sensitiveurl: {0}'.format(config.getoption('sensitive_url'))
+    base_url = config.getoption('base_url')
+    sensitive_url = config.getoption('sensitive_url')
+    msg = 'sensitiveurl: {0}'.format(config.getoption('sensitive_url'))
+    if base_url and sensitive_url and re.match(sensitive_url, base_url):
+        msg += '\nwarning: base URL ({}) matches sensitive URL ({})'\
+            .format(base_url, sensitive_url)
+    return msg
 
 
 @pytest.fixture(scope='session')


pytest_selenium/safety.py
```diff
@@ -63,8 +63,9 @@ def _skip_sensitive(request, sensitive_url):
...
-        pytest.skip(
-            'This test is destructive and the target URL is '
-            'considered a sensitive environment. If this test is '
-            'not destructive, add the \'nondestructive\' marker to '
-            'it. Sensitive URL: {0}'.format(sensitive_url))
+        message = 'This test is destructive and the target URL is '\
+            'considered a sensitive environment. If this test is '\
+            'not destructive, add the \'nondestructive\' marker to '\
+            'it. Sensitive URL: {0}'.format(sensitive_url)
+        print(message)
```
BeyondEvil: @Jenselme If you want it to print w/o using `-s`, you can import `warning` and do `warning.warn(message)`

Jenselme: @BeyondEvil just tested with `warnings.warn` and nothing is logged without the `-s` option (there is no module named `warning`).

BeyondEvil: @Jenselme Sorry, the module is called `warnings`
https://docs.python.org/2/library/warnings.html

Jenselme: @BeyondEvil as I said in my previous comment, output of `warnings.warn` is swallowed too. 3 travis jobs didn't start, that why travis is unhappy.

docs/user_guide.rst
```diff
@@ -43,6 +43,10 @@ Specifying a Base URL
 
 To specify a base URL, refer to the documentation for the
 `pytest-base-url <https://github.com/pytest-dev/pytest-base-url>`_ plugin.
+Don't forget to configure the :ref:`sensitive environments <sensitive-environments>`
+or your tests will be skipped.
```
davehunt: Only destructive tests that use the base URL fixture will be skipped when the base URL matches a sensitive environment. I think it's important to make this clear. Perhaps reword this as:

```rst
Note: By default, any tests using a base URL will be skipped. This is because all tests
are considered destructive, and all environments are considered sensitive. See
`Sensitive Environments`_ for further details.
```

docs/user_guide.rst
```diff
@@ -43,6 +43,10 @@ Specifying a Base URL
...
+
+.. _sensitive-environments:
```
davehunt: There's no need for an explicit link here, and all titles are considered links.

pytest_selenium/safety.py
davehunt: I'm not keen on adding a print statement. As you highlight, this will not be shown by default anyway. The tests will be skipped, and to see details of skipped tests the user is expected to add `-rs` to the command line as documented [here](https://docs.pytest.org/en/latest/skipping.html).

I would be okay with adding a single visible warning before the tests if the base URL matches the sensitive URL.

davehunt: FWIW if you're using [pytest-html](https://github.com/pytest-dev/pytest-html) the skip reasons are always included in the report.

Jenselme: > FWIW if you're using pytest-html the skip reasons are always included in the report.

Good to know, thanks. I'll try to add the warning before the tests today.

Jenselme:
@davehunt PR updated. Sorry for the delay, I was quite busy these days. Some tests failed on travis and I don't understand why it is related to my modifications.

davehunt:
Thanks @Jenselme, I rebased, merged, and added a test for the new warning. I have a few bugs to look into, but will hopefully get a new release out in the next few days.

Jenselme:
@davehunt thanks. Looking forward to it!

davehunt:
@Jenselme it's released! https://pypi.python.org/pypi/pytest-selenium/1.11.0

Jenselme:
@davehunt thanks!

PR #127, opened 2017-06-03
Update remote.py

oliviertassinari:


davehunt:
Thanks for the pull request @oliviertassinari! Unfortunately, this causes a failure because the `browser_profile` keyword argument is not passed to the driver class. As Firefox is the only remote browser to use this keyword argument, and the preferred method of providing a profile is via `firefox_options`, I've decided not to merge this patch.

oliviertassinari:
Sounds fair :+1:.

PR #137, opened 2017-11-06
Fix capabilities issues with latest Selenium server

### LINKED ISSUE #136: Remote Firefox driver passes incorrect capabilities
Using the latest 3.7.0 docker selenium firefox image and pytest-selenium 1.11.1 I get an error that states ```unable to find matching set of capabilities```. 

I removed pytest selenium and followed the remote driver instructions found here: http://selenium-python.readthedocs.io/getting-started.html#using-selenium-with-remote-webdriver and I was able to get the tests to pass again.

I am thinking the remote capabilities that pytest-selenium copies has been outdated, or maybe the selenium has changed the data type.

davehunt:
Fixes #136, which was caused by the default capability value of 'ANY'. This may be a regression in Selenium, but the value is superfluous anyway. By not specifying version or platform, they will not be used when matching with a node, which is effectively the same as 'ANY'. I also removed the 'browserName' check as I'd prefer not to include any assumptions of the remove server's supported capabilities. If for instance you have a remote server only providing a single browser/platform, then you should be able to omit all capabilities for matching.


pytest_selenium/pytest_selenium.py
```diff
@@ -48,11 +55,20 @@ def capabilities(request, driver_class, chrome_options, firefox_options,
...
     capabilities = copy.deepcopy(session_capabilities)  # make a copy
     if driver_class == webdriver.Remote:
-        browser_name = str(capabilities.get('browserName')).lower()
-        if browser_name == 'chrome':
-            capabilities.update(chrome_options.to_capabilities())
-        elif browser_name == 'firefox':
-            capabilities.update(firefox_options.to_capabilities())
+        browser = capabilities.get('browserName', '').upper()
+        key, options = (None, None)
+        if browser == 'CHROME':
```
BeyondEvil:
```        if browser == 'CHROME':
            key = 'chromeOptions'
            options = chrome_options.to_capabilities()
            if 'chromeOptions' not in options:
                key = 'goog:chromeOptions'
        elif browser == 'FIREFOX':```

davehunt: Good suggestion, though I tweaked it a little further.

davehunt:
Merged.

pytest_selenium/pytest_selenium.py
BeyondEvil: @davehunt But isn't `goog:chromeOptions` the key that is being replaced by just `chromeOptions`?

PR #140, opened 2017-11-21
Be more lenient with driver allowing lowercase

BeyondEvil:
Allow lowercase when defining what `--driver` to use.

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -8,6 +8,7 @@ from datetime import datetime
 import os
 
 import pytest
+from requests.structures import CaseInsensitiveDict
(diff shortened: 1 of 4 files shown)


pytest_selenium/pytest_selenium.py
```diff
@@ -108,7 +109,8 @@ def driver_class(request):
     driver = request.config.getoption('driver')
     if driver is None:
         raise pytest.UsageError('--driver must be specified')
-    return getattr(webdriver, driver, webdriver.Remote)
+    driver = SUPPORTED_DRIVERS[driver.lower()]
+    return driver
```
davehunt: Rather redefine the `driver` variable, could we just return the class directly?

PR #143, opened 2017-11-30
Fix a DeprecationWarning on selenium >= 3.8.0.

D3X:
Selenium 3.8.0 has deprecated `chrome_options`, so we should use `options` instead.


davehunt:
@D3X could you do the same for Firefox, which has the same deprecation? In the future once this has been removed I think we can make 3.8 our minimum supported version and remove this logic.

D3X:
@davehunt done!

davehunt:
Thanks, I'm just rerunning the Travis job now that I've fixed the flake8 issue. I'm hoping to tag a release tomorrow.

Issue #142, opened 2017-11-29
Order of remote driver arguments


micheletest:
I recently updated python packages, including this one, and ran across a potential issue with the remote driver. If I run the pytest command with `--host notlocalhost --driver Remote --port 1234` (just an example) it does not work. If I run the command with `--driver Remote --port 1234 --host notlocalhost` it does work. For the command with the host first, the host seems to become the default `localhost` instead of the provided host `notlocalhost`.

The actual error due to the incorrect host is: `E urllib.error.URLError: <urlopen error [Errno 111] Connection refused>`

I updated pytest-selenium from 1.9.1 to 1.11.2. Using 1.9.1 I did not see this error.

davehunt:
Interesting. It's possible that we need to avoid overwriting the host/port if they've been set in advance of the driver:

https://github.com/pytest-dev/pytest-selenium/blob/52b9b71197994233f6a5d2bca8a221f5c535690b/pytest_selenium/pytest_selenium.py#L292-L293

micheletest:
Yes, I think that would do it - don't do anything if there is already a host or port. Otherwise use the existing logic. 

```        # set the default host and port if not specified already and if specified in the driver module
        if not namespace.host:
            setattr(namespace, 'host', getattr(driver, 'HOST', None))
        if not namespace.port:
            setattr(namespace, 'port', getattr(driver, 'PORT', None))
```

PR #145, opened 2017-11-30
Fix command line argument ordering issue

BeyondEvil:
Here's my suggested solution for  #142  @davehunt 

~~Sorry, I failed on the test part. Tried using metadata, but no dice.~~

Made a best effort at some tests...


pytest_selenium/pytest_selenium.py
```diff
@@ -94,8 +94,12 @@ def driver_kwargs(request, capabilities, chrome_options, driver_args,
         driver_path=driver_path,
         firefox_options=firefox_options,
         firefox_profile=firefox_profile,
-        host=pytestconfig.getoption('host'),
-        port=pytestconfig.getoption('port'),
+        host=pytestconfig.getoption('host',
+                                    getattr(driver, 'HOST', None),
+                                    True),
```
davehunt: According to the source code for `getoption`, passing `skip=True` would "raise pytest.skip if option does not exists or has a None value". Is this what we want?

pytest_selenium/pytest_selenium.py
```diff
@@ -287,10 +291,6 @@ class DriverAction(argparse.Action):
 
     def __call__(self, parser, namespace, values, option_string=None):
         setattr(namespace, self.dest, values)
-        driver = getattr(drivers, values.lower())
-        # set the default host and port if specified in the driver module
-        setattr(namespace, 'host', getattr(driver, 'HOST', None))
-        setattr(namespace, 'port', getattr(driver, 'PORT', None))
```
davehunt: I think this means we can remove the custom action entirely?

testing/test_driver.py
```diff
@@ -52,3 +52,45 @@ def test_driver_quit(selenium):
...
+
+def test_default_host_port(testdir):
+    host = 'localhost'
+    port = 4444
+    file_test = testdir.makepyfile("""
+        import pytest
+        @pytest.mark.nondestructive
+        def test_pass(driver_kwargs):
+            assert driver_kwargs['command_executor'] == 'http://{}:{}/wd/hub'
+    """.format(host, port))
```
davehunt: Rather than checking command_executor, can we not check host and port arguments directly?

testing/test_driver.py
```diff
@@ -52,3 +52,45 @@ def test_driver_quit(selenium):
...
+    testdir.quick_qa('--host', host,
+                     '--driver', 'Remote',
+                     '--port', port,
+                     file_test, passed=1)
```
davehunt: Have you checked that this fails without your fix?

pytest_selenium/pytest_selenium.py
BeyondEvil: Yes, because we are providing a default, even if that default may be `None`.
Not setting `skip` truthy will then not give us the default. At least that's what happened to me.

pytest_selenium/pytest_selenium.py
BeyondEvil: Probably, didn't think to try it.

testing/test_driver.py
BeyondEvil: That was my initial implementation, but I got key-error on the dict and on inspection the only key it had with the host/port values were command_executor.

testing/test_driver.py
BeyondEvil: Yes. :)

pytest_selenium/pytest_selenium.py
```diff
@@ -94,8 +94,12 @@ def driver_kwargs(request, capabilities, chrome_options, driver_args,
...
         firefox_options=firefox_options,
         firefox_profile=firefox_profile,
-        host=pytestconfig.getoption('host'),
-        port=pytestconfig.getoption('port'),
+        host=pytestconfig.getoption('host',
+                                    getattr(driver, 'HOST', None),
+                                    True),
+        port=pytestconfig.getoption('port',
+                                    getattr(driver, 'PORT', None),
+                                    True),
```
davehunt: Doing it this way means the `pytest_configure` hook isn't able to access these values, which it uses for adding to the metadata. I'm going to see if we can simply move this code into the hook.

davehunt:
Merged in https://github.com/pytest-dev/pytest-selenium/commit/453a0e3685e83f3aebc33dff655f70c0af767a46

PR #147, opened 2017-12-07
Implementation of SauceLabs job auth options

BeyondEvil:
Just the implementation. Will add docs in the next commit.

As far as tests goes. Is it even possible? It would require me to access the `summary` from the `pytest_selenium_runtest_makereport` hook, and check for the presence of a correct url. At least for the `none` and `token` (default) options.

### DIFF
--- a/pytest_selenium/drivers/saucelabs.py
+++ b/pytest_selenium/drivers/saucelabs.py
@@ -19,7 +19,7 @@ class SauceLabs(Provider):
 
     @property
     def auth(self):
-        return (self.username, self.key)
+        return self.username, self.key
 
     @property
     def executor(self):
@@ -62,7 +62,7 @@ def pytest_selenium_runtest_makereport(item, report, summary, extra):
 
     # Add the job URL to the summary
     provider = SauceLabs()
-    job_url = provider.JOB.format(session=session_id)
+    job_url = get_job_url(item.config, provider, session_id)
     summary.append('{0} Job: {1}'.format(provider.name, job_url))
     pytest_html = item.config.pluginmanager.getplugin('html')
     # Add the job URL to the HTML report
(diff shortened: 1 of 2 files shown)


pytest_selenium/drivers/saucelabs.py
```diff
@@ -143,3 +143,38 @@ def _video_html(session):
...
+        key = '{0.username}:{0.key}:{1}'.format(provider, ttl)
+
+    return get_auth_url(job_url, key, session_id)
+
+
+def get_auth_url(url, key, session_id):
+    import hmac
+    from hashlib import md5
+
+    auth_url = url.format(session='{}?auth={}')
```
davehunt: Rather than forcing the parameter into the `session`, you could create a new string based on the resolved job URL (with the session) and the token.

pytest_selenium/drivers/saucelabs.py
```diff
@@ -143,3 +143,38 @@ def _video_html(session):
...
+    auth_token = hmac.new(key.encode('utf-8'),
+                          session_id.encode('utf-8'),
+                          md5).hexdigest()
+    return auth_url.format(session_id, auth_token)
```
davehunt: So this would be:

```python
return '{}?auth={}'.format(url, token)
```

pytest_selenium/pytest_selenium.py
```diff
@@ -303,6 +303,12 @@ def pytest_addoption(parser):
                   help='debug to exclude from capture',
                   default=os.getenv('SELENIUM_EXCLUDE_DEBUG'))
 
+    _auth_choices = ('none', 'token', 'hour', 'day')
+    parser.addini('saucelabs_job_auth',
+                  help='Authorization for the SauceLabs job {0}'.
```
davehunt: nit: Sauce Labs

pytest_selenium/pytest_selenium.py
```diff
@@ -303,6 +303,12 @@ def pytest_addoption(parser):
                   help='debug to exclude from capture',
                   default=os.getenv('SELENIUM_EXCLUDE_DEBUG'))
 
+    _auth_choices = ('none', 'token', 'hour', 'day')
```
davehunt: I'd prefer these values to be in upper case.

pytest_selenium/pytest_selenium.py
davehunt: Also, could we make it more clear that these are the options.

pytest_selenium/pytest_selenium.py
```diff
@@ -303,6 +303,12 @@ def pytest_addoption(parser):
...
+    parser.addini('saucelabs_job_auth',
+                  help='Authorization for the SauceLabs job {0}'.
+                  format(_auth_choices),
+                  default=os.getenv('SAUCELABS_JOB_AUTH', 'token'))
```
davehunt: I feel that the default should be NONE, for two reasons. The first is that this means there is no change of behaviour when upgrading, and the second is that I don't feel that we should be generating permanent public links to jobs by default.

pytest_selenium/pytest_selenium.py
BeyondEvil: Should I also make the values for `_capture_choices` upper case? @davehunt

PR #153, opened 2018-02-05
Added changes to allow driver log to always be added to the HTML report.

### LINKED ISSUE #149: Add driver log to HTML report even if there is no driver instance
We currently only add the driver log to the HTML report if there is a driver instance, however this means we fail to include the logs when there is an issue actually creating the driver instance.

We should move the following lines from `_gather_logs`:

https://github.com/pytest-dev/pytest-selenium/blob/2c40b5c395162f598ed02e7302d3ed4d71a0873b/pytest_selenium/pytest_selenium.py#L245-L249

To outside the `if driver is not None` in `pytest_runtest_makereport`:

https://github.com/pytest-dev/pytest-selenium/blob/2c40b5c395162f598ed02e7302d3ed4d71a0873b/pytest_selenium/pytest_selenium.py#L186

This should help with diagnosing issues when staring the driver instance, such as in https://github.com/mozilla/FoxPuppet/pull/115

jrbenny35:
Fixes #149

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -183,9 +183,13 @@ def pytest_runtest_makereport(item, call):
     failure = (report.skipped and xfail) or (report.failed and not xfail)
     when = item.config.getini('selenium_capture_debug').lower()
     capture_debug = when == 'always' or (when == 'failure' and failure)
-    if driver is not None:
-        if capture_debug:
-            exclude = item.config.getini('selenium_exclude_debug').lower()
+    if capture_debug:
+        exclude = item.config.getini('selenium_exclude_debug').lower()
+        if 'logs' not in exclude:
+            # gather logs that do not depend on a driver instance
+            _gather_driver_log(item, summary, extra)
+        if driver is not None:
+            # gather debug that depends on a driver instance
             if 'url' not in exclude:
                 _gather_url(item, report, driver, summary, extra)
             if 'screenshot' not in exclude:
@@ -194,12 +198,13 @@ def pytest_runtest_makereport(item, call):
                 _gather_html(item, report, driver, summary, extra)
             if 'logs' not in exclude:
                 _gather_logs(item, report, driver, summary, extra)
+            # gather debug from hook implementations
             item.config.hook.pytest_selenium_capture_debug(
                 item=item, report=report, extra=extra)
+    if driver is not None:
+        # allow hook implementations to further modify the report
         item.config.hook.pytest_selenium_runtest_makereport(
             item=item, report=report, summary=summary, extra=extra)
-    # always gather driver logs
-    _gather_driver_logs(item, summary, extra)
     if summary:
         report.sections.append(('pytest-selenium', '\n'.join(summary)))
     report.extra = extra
@@ -262,7 +267,7 @@ def _gather_logs(item, report, driver, summary, extra):
                 format_log(log), '%s Log' % name.title()))
 
 
-def _gather_driver_logs(item, summary, extra):
+def _gather_driver_log(item, summary, extra):
     pytest_html = item.config.pluginmanager.getplugin('html')
     if hasattr(item.config, '_driver_log') and \
        os.path.exists(item.config._driver_log):
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -198,6 +198,8 @@ def pytest_runtest_makereport(item, call):
                 item=item, report=report, extra=extra)
         item.config.hook.pytest_selenium_runtest_makereport(
             item=item, report=report, summary=summary, extra=extra)
+    # always gather driver logs
+    _gather_driver_logs(item, summary, extra)
     if summary:
         report.sections.append(('pytest-selenium', '\n'.join(summary)))
     report.extra = extra
@@ -242,11 +244,6 @@ def _gather_html(item, report, driver, summary, extra):
 
 def _gather_logs(item, report, driver, summary, extra):
     pytest_html = item.config.pluginmanager.getplugin('html')
-    if item.config._driver_log and os.path.exists(item.config._driver_log):
-        if pytest_html is not None:
-            with io.open(item.config._driver_log, 'r', encoding='utf8') as f:
-                extra.append(pytest_html.extras.text(f.read(), 'Driver Log'))
-        summary.append('Driver log: {0}'.format(item.config._driver_log))
     try:
         types = driver.log_types
     except Exception as e:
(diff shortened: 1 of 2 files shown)


pytest_selenium/pytest_selenium.py
```diff
@@ -198,6 +198,8 @@ def pytest_runtest_makereport(item, call):
                 item=item, report=report, extra=extra)
         item.config.hook.pytest_selenium_runtest_makereport(
             item=item, report=report, summary=summary, extra=extra)
+    # always gather driver logs
+    _gather_driver_logs(item, summary, extra)
```
davehunt: We should only do this if `capture_debug` is True. Perhaps swap the `driver is not None` and `capture_debug` if blocks around.

pytest_selenium/pytest_selenium.py
```diff
@@ -265,6 +262,27 @@ def _gather_logs(item, report, driver, summary, extra):
                 format_log(log), '%s Log' % name.title()))
 
 
+def _gather_driver_logs(item, summary, extra):
+    pytest_html = item.config.pluginmanager.getplugin('html')
+    try:
+        item.config._driver_log
+    except Exception:
+        summary.append('No Driver Log found')
```
davehunt: Rather than try except, let's just use `if item.config._driver_log and os.path.exists(item.config._driver_log)`

jrbenny35: With this, if ```item.config._driver_log``` doesn't exist, it returns ```None``` and then ```os.path.exists``` tries to find that and throws a NoneType error.

jrbenny35:
I looked at the log tests but maybe I could work on that after this?

pytest_selenium/pytest_selenium.py
davehunt: If `item.config._driver_log` is `None` then it is not truthy, so the following condition should not be checked.

davehunt:
> I looked at the log tests but maybe I could work on that after this?

Sure, though it may just be a case of removing comments from the tests.

pytest_selenium/pytest_selenium.py
jrbenny35: Hmm, it seemed like that wasn't happening but I'll check again.

pytest_selenium/pytest_selenium.py
```diff
@@ -194,10 +194,12 @@ def pytest_runtest_makereport(item, call):
                 _gather_html(item, report, driver, summary, extra)
             if 'logs' not in exclude:
                 _gather_logs(item, report, driver, summary, extra)
+            item.config.hook.pytest_selenium_runtest_makereport(
+                item=item, report=report, summary=summary, extra=extra)
```
davehunt: I believe this needs to be outside `if capture_debug` as it is a hook for other plugins to add details to the HTML report. They may want to do their own check for if debug information is desired, or enhance the report regardless of this condition. It's probably safe to do this outside the `driver is not None` check, as this can also be checked by anyone implementing the hook.

pytest_selenium/pytest_selenium.py
```diff
@@ -265,6 +262,16 @@ def _gather_logs(item, report, driver, summary, extra):
                 format_log(log), '%s Log' % name.title()))
 
 
+def _gather_driver_logs(item, summary, extra):
+    pytest_html = item.config.pluginmanager.getplugin('html')
+    if hasattr(item.config, '_driver_log') and \
+       os.path.exists(item.config._driver_log):
```
davehunt: This is interesting. Was `item.config._driver_log` not `None`? I would actually expect this to fail with the details from your initial comment, as a `None` value here would pass the first condition and then fail the second.

jrbenny35: I was getting an attribute error.

pytest_selenium/pytest_selenium.py
davehunt: This is back to the original issue. We should only gather the driver logs if `capture_debug` is True. We just also need to call `item.config.hook.pytest_selenium_runtest_makereport` regardless of the value of `capture_debug`.

pytest_selenium/pytest_selenium.py
davehunt: We explicitly set the `_driver_log` attribute here: https://github.com/pytest-dev/pytest-selenium/blob/a90604245a1902f320e2caecbf2187e20355bac1/pytest_selenium/pytest_selenium.py#L104

This means it should be save to check the value rather than the presence. Can you provide the traceback for the error you were seeing?

pytest_selenium/pytest_selenium.py
jrbenny35: Is that why you said to switch the order of them?

davehunt: Yeah, I just thought that would be cleaner. At the time I was forgetting about the hook, but I think this is fine outside of the driver check. Something like:

```python
if capture_debug:
    _gather_driver_logs(item, summary, extra)
    if driver is not None:
        # selenium related debug capturing
item.config.hook.pytest_selenium_runtest_makereport(
    item=item, report=report, summary=summary, extra=extra)

Comment (jrbenny35) [reply pytest_selenium/pytest_selenium.py]: I'll have to look over it again, I think I was getting an error with that.

Comment (jrbenny35) [reply pytest_selenium/pytest_selenium.py]: Here is a log: https://pastebin.mozilla.org/9077828

**Comment by jrbenny35** [comment]
Looking at this a bit deeper, if we switch ```capture_debug``` and the ```driver``` checks around, but move the ```pytest_selenium_runtest_makereport``` outside of capture debug, I think we will always get the error ```AttributeError: 'Function' object has no attribute '_driver'``` as it seems this hook is expecting there to be a driver. The current code has this makereport within the ```driver``` check.

**Comment by davehunt** [comment]
Sorry for dropping the ball here, I'll squash your commits and push a followup.

**Comment by davehunt** [comment]
@jrbenny35 let me know what you think of these changes. Thanks!

File: pytest_selenium/pytest_selenium.py [hunk 5]
```diff
@@ -194,8 +198,11 @@ def pytest_runtest_makereport(item, call):
                 _gather_html(item, report, driver, summary, extra)
             if 'logs' not in exclude:
                 _gather_logs(item, report, driver, summary, extra)
+            # gather debug from hook implementations
             item.config.hook.pytest_selenium_capture_debug(
                 item=item, report=report, extra=extra)
+    if driver is not None:
```
jrbenny35: Ah you are checking the driver twice. I was thinking it was only needing to be done once.

jrbenny35:
Looks good dave! r+

pytest_selenium/pytest_selenium.py
davehunt: Yeah, this seemed the least complicated approach.

PR #163, opened 2018-04-27
Fix failing test using conditional

BeyondEvil:


pytest_selenium/pytest_selenium.py
```diff
@@ -43,8 +43,8 @@ def pytest_addhooks(pluginmanager):
 def session_capabilities(pytestconfig):
     """Returns combined capabilities from pytest-variables and command line"""
     driver = pytestconfig.getoption('driver').upper()
-    capabilities = getattr(DesiredCapabilities, driver, {}).copy()
-    if driver == 'REMOTE':
+    capabilities = getattr(DesiredCapabilities, driver.upper(), {}).copy()
+    if driver.upper() == 'REMOTE':
```
davehunt: driver is already converted to uppercase on line 45, so we can remove these changes

BeyondEvil: *the feeling of shame*

PR #165, opened 2018-05-18
Use https, rather than http for API and job URLs

stephendonner:
All the API/job, etc. URLs I've seen on Sauce Labs' documentation use HTTPS; I'm not sure if this is the right change, but it seems safe, and preferred?

https://wiki.saucelabs.com/display/DOCS/Accessing+the+API

@davehunt r?

### DIFF
--- a/pytest_selenium/drivers/saucelabs.py
+++ b/pytest_selenium/drivers/saucelabs.py
@@ -14,8 +14,8 @@ from pytest_selenium.drivers.cloud import Provider
 
 class SauceLabs(Provider):
 
-    API = 'http://saucelabs.com/rest/v1/{username}/jobs/{session}'
-    JOB = 'http://saucelabs.com/jobs/{session}'
+    API = 'https://saucelabs.com/rest/v1/{username}/jobs/{session}'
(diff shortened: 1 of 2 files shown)


stephendonner:
Er, sorry; just realized doing SSL/TLS over port 80 (by default) won't work, sigh.

pytest_selenium/drivers/saucelabs.py
```diff
@@ -14,16 +14,16 @@
...
+    JOB = 'https://saucelabs.com/jobs/{session}'
 
     @property
     def auth(self):
         return (self.username, self.key)
 
     @property
     def executor(self):
-        return 'http://{0}:{1}@ondemand.saucelabs.com:80/wd/hub'.format(
+        return 'https://{0}:{1}@ondemand.saucelabs.com:80/wd/hub'.format(
```
davehunt: You should remove the hard-coded port here.

stephendonner:
I took a stab at a few other provider updates in https://github.com/pytest-dev/pytest-selenium/pull/167, but it needs a bit of work.

PR #168, opened 2018-06-08
updated auth token test

BeyondEvil:
Let me know what you think @davehunt 👍


davehunt:
I'm wondering if there's value in trying to verify the token part, especially with the mocking. Perhaps it would be enough to verify the day/hour feature and ignore the token?

BeyondEvil:
You're probably right @davehunt, but could clarify what you mean with verifying the day/hour part?

davehunt:
So we could just leave this as it is, beause I believe it's verifying everything we need it to. I guess what I was thinking was that we could simplify it to two tests: `test_auth_type_none` and `test_auth_type_expiration`. The first would be a simple check that the URL is not modified, and the second could be parameterised with the range and date/time format. The test could verify that the URL matches a regular expression, which ignores the token value entirely rather than mocking it out.

BeyondEvil:
Ooh, sounds like fun @davehunt I'll see what I can accomplish. Simple/easy is always good. 👍

BeyondEvil:
Something like this @davehunt ?

davehunt:
Looks good, thanks @BeyondEvil

davehunt:
Merged in https://github.com/pytest-dev/pytest-selenium/commit/bea3f44b67c45cdeb15251e32eca0e0331c542b5

PR #169, opened 2018-06-28
Fix provider bug

BeyondEvil:
Fix for issue discussed on Slack (+tests)


pytest_selenium/drivers/browserstack.py
```diff
@@ -14,13 +14,17 @@ class BrowserStack(Provider):
...
+        return self.username, self.key
 
     @property
     def executor(self):
         return 'http://{0}:{1}@hub.browserstack.com:80/wd/hub'.format(
             self.username, self.key)
 
+    @property
+    def name(self):
+        return type(self).__name__
```
davehunt: We can just inherit this from `Provider`

pytest_selenium/drivers/cloud.py
```diff
@@ -17,15 +17,15 @@ class Provider(object):
 
     @property
     def driver(self):
-        return type(self).__name__
+        return type(self).__name__.lower()
 
     @property
     def name(self):
         return self.driver
```
davehunt: Can we switch this around so this returns `type(self).__name__` and `driver` returns `self.name.lower()`?

testing/test_driver.py
```diff
@@ -115,3 +115,37 @@ def test_pass(driver_kwargs):
...
+        import pytest
+        @pytest.mark.nondestructive
+        def test_pass(request):
+            assert request.config.getoption('driver') == 'saucelabs'
+    """)
+    testdir.quick_qa('--driver', 'SauceLabs',
+                     file_test, passed=1)
+
+
+def test_provider_lowercase():
```
davehunt: Let's parameterise this test.

BeyondEvil: I thought about it, but I'm not sure of how to accomplish that with the imports... 🤔

pytest_selenium/drivers/browserstack.py
BeyondEvil: Yeah, that's the way it was, but then the tests failed. But if we do your suggestion below and the `uses_driver()` function, we can restore it. 👍

pytest_selenium/drivers/cloud.py
BeyondEvil: Agreed.

testing/test_driver.py
BeyondEvil: yeah, I need help with this @davehunt  (the imports).

pytest_selenium/drivers/saucelabs.py
```diff
@@ -27,6 +27,10 @@ def executor(self):
         return 'https://{0}:{1}@ondemand.saucelabs.com/wd/hub'.format(
             self.username, self.key)
 
+    @property
+    def driver(self):
+        return 'saucelabs'
```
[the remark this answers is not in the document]
BeyondEvil: We can (incl. the entire `driver` property), but to do that, one of two other things must change.

1. "Sauce Labs" needs to become "SauceLabs". That is, removing the overloaded `name` property in the `SauceLabs` class.

2. `uses_driver` needs to be overloaded to remove the space.

```
    def uses_driver(self, driver):
        return driver.lower() == self.name.lower().replace(' ', '')
```

testing/test_driver.py
```diff
@@ -115,3 +118,14 @@ def test_pass(driver_kwargs):
...
+
+@pytest.mark.parametrize(('module_name', 'class_name'),
+                         [(saucelabs, 'Sauce Labs'),
+                          (testingbot, 'TestingBot'),
+                          (crossbrowsertesting, 'CrossBrowserTesting'),
+                          (browserstack, 'BrowserStack')])
+def test_provider_naming(module_name, class_name):
+    provider = getattr(module_name, class_name.replace(' ', ''))()
+    assert provider.uses_driver(class_name.replace(' ', ''))
+    assert provider.name == class_name
```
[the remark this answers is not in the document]
BeyondEvil: Ah, nice!

pytest_selenium/drivers/cloud.py
```diff
@@ -41,3 +42,6 @@ def get_credential(self, key, envs):
                 if value:
                     return value
         raise MissingCloudCredentialError(self.name, key, envs)
+
+    def uses_driver(self, driver):
+        return driver.lower() == self.driver
```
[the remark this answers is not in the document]
BeyondEvil: Yes, afaict the `driver` property is only used in one other place, and that is `config()` (for getting the dot-configuration files), and that can use the `name` property just as well.

pytest_selenium/drivers/saucelabs.py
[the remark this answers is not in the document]
BeyondEvil: Let's go with option 1 @davehunt

PR #170, opened 2018-06-28
Use a GitHub auth token for determining the latest version of GeckoDriver

davehunt:


davehunt:
It failed, but I actually think this will fail for any builds not originating from this repository. Let's just hardcode teh version.

PR #172, opened 2018-06-30
Fix invalid escape sequences

asottile:


davehunt:
@asottile could you explain why these changes are needed and what they're fixing?

testing/test_destructive.py
```diff
@@ -13,7 +13,7 @@ def test_skip_destructive_by_default(testdir):
 
 
 def test_warn_when_url_is_sensitive(testdir, httpserver, monkeypatch, capsys):
-    monkeypatch.setenv('SENSITIVE_URL', '127\.0\.0\.1')
+    monkeypatch.setenv('SENSITIVE_URL', r'127\.0\.0\.1')
```
davehunt: @asottile can you explain in more detail? This is a string that's later treated as a regular expression.. Wouldn't making this a raw string literal mean the backslashes would need escaping?

asottile: This is a noop in versions of python that don't mark this as a syntax error (py38+) (previously unknown escape sequences preserved the back slashes) - - in a raw literal they are explicitly preserved. A common misconception is that the r prefix is for regex, though it _is_ the most common use of raw literals in practice.

PR #173, opened 2018-07-03
Run tests against Python 3.7 and update classifiers

davehunt:


asottile:
Note that travis-ci py37 is 3.7a4 which is pretty different from 3.7.0 final

For more info : https://github.com/deadsnakes/travis-ci-python3.7-example

davehunt:
Thanks @asottile I think I'll put this on ice for now then.

asottile:
there is a not-so-terrible option now with `dist: xenial` -- I just updated the README: https://github.com/deadsnakes/travis-ci-python3.7-example#update-2018-07-03

PR #177, opened 2018-07-28
Moved username and key to capabilities + tests

BeyondEvil:
Moved Sauce specific tests from `test_capabilities.py` to `test_saucelabs.py` as well.


testing/test_saucelabs.py
```diff
@@ -77,6 +78,83 @@ def test_invalid_credentials_file(failure, monkeypatch, tmpdir):
...
+    capabilities = {'browserName': 'chrome'}
+    variables = testdir.makefile('.json', '{{"capabilities": {}}}'.format(
+        json.dumps(capabilities)))
+
+    file_test = testdir.makepyfile("""
+        import pytest
+        @pytest.mark.nondestructive
+        def test_sauce_capabilities(driver_kwargs):
+            assert driver_kwargs['desired_capabilities']['username'] == '{}'
+            assert driver_kwargs['desired_capabilities']['accessKey'] == '{}'
```
davehunt: The purpose of this test is to check that `sauce:options` is not present, so I don't think we need to check the capabilities here.

davehunt:
@BeyondEvil I'm happy for you to merge this with or without my comments addressed. Thanks!

testing/test_saucelabs.py
BeyondEvil: Dammit, I should've followed my instincts. Will move them out to their own test. 👍 

Will add a section to the release-notes as well. :)

docs/news.rst
```diff
@@ -1,6 +1,14 @@
 Release Notes
 =============
 
+
+TBD
```
davehunt: We can assume it will be 1.14.0 and I typically just set the date as "(unreleased)".

PR #178, opened 2018-08-08
Use HTTPS where we can, and remove hard-coded port 80

stephendonner:
Sorry for the delay, and the mess.  I've taken another clean shot (hopefully) at this.

### DIFF
--- a/pytest_selenium/drivers/browserstack.py
+++ b/pytest_selenium/drivers/browserstack.py
@@ -18,7 +18,7 @@ class BrowserStack(Provider):
 
     @property
     def executor(self):
-        return 'http://{0}:{1}@hub.browserstack.com:80/wd/hub'.format(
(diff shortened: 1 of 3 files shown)


BeyondEvil:
Do all of the providers support HTTPS? Have you tested this? I can test against Testingbot if need be, because I have an account there. @stephendonner

stephendonner:
@BeyondEvil thanks; I could clearly use the help, here.  Yes, TestingBot's API supports HTTPS, and is used in their API docs' examples: https://testingbot.com/support/api

The Python client, though, uses primarily HTTP throughout its docs.

BeyondEvil:
This should now be handled by #190 #191 and #192

BeyondEvil:
As my previous comment stated, this has now been handled.
You can go head an close this PR if there's nothing else you want to add @stephendonner

Thanks for bringing this to us.   👍

PR #180, opened 2018-08-13
remove log_path and use service_log_path consistently instead

### LINKED ISSUE #179: DeprecationWarning: use service_log_path instead of log_path
Hi!

In the new v3.14.0 release of the Python selenium package, a new deprecation warning has been added, which is currently triggered by pytest-selenium:
```
...
    def driver(request, driver_class, driver_kwargs):
        """Returns a WebDriver instance based on options and capabilities"""
>       driver = driver_class(**driver_kwargs)
../../../venv/lib/python2.7/site-packages/pytest_selenium/pytest_selenium.py:154: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
...
    def __init__(self, firefox_profile=None, firefox_binary=None,
                 timeout=30, capabilities=None, proxy=None,
                 executable_path="geckodriver", options=None,
                 service_log_path="geckodriver.log", firefox_options=None,
                 service_args=None, desired_capabilities=None, log_path=None):
...
        if log_path:
>           warnings.warn('use service_log_path instead of log_path', DeprecationWarning)
E           DeprecationWarning: use service_log_path instead of log_path
../../../venv/lib/python2.7/site-packages/selenium/webdriver/firefox/webdriver.py:104: DeprecationWarning
```

See:
https://github.com/SeleniumHQ/selenium/pull/5979
https://travis-ci.org/mozilla/treeherder/jobs/411325051#L947

BeyondEvil:
Fixes #179


testing/conftest.py
```diff
@@ -24,7 +24,7 @@ def testdir(request, httpserver_base_url):
     if 'testdir' not in item.funcargnames:
         return
 
-    testdir = request.getfuncargvalue('testdir')
+    testdir = request.getfixturevalue('testdir')
```
BeyondEvil: This was caught when running with DeprecationWarning enabled. @davehunt

testing/test_crossbrowsertesting.py
```diff
@@ -59,11 +59,15 @@ def test_invalid_credentials_env(failure, monkeypatch, tmpdir, username, key):
...
 
 
 def test_invalid_credentials_file(failure, monkeypatch, tmpdir):
     monkeypatch.setattr(os.path, 'expanduser', lambda p: str(tmpdir))
     config = tmpdir.join('.crossbrowsertesting')
     config.write('[credentials]\nusername=foo\nkey=bar')
-    failure('--capability', 'browser_api_name', 'FF46')
+    out = failure('--capability', 'browser_api_name', 'FF46')
+    messages = ['missing auth', 'basic auth failed']
+    assert any(message in out for message in messages)
```
BeyondEvil: I added this because the test was passing for the wrong reason.

testing/test_testingbot.py
```diff
@@ -75,4 +75,4 @@ def test_invalid_host(failure, monkeypatch, tmpdir):
     monkeypatch.setattr(os.path, 'expanduser', lambda p: str(tmpdir))
     tmpdir.join('.testingbot').write('[credentials]\nkey=foo\nsecret=bar')
     out = failure('--host', 'foo.bar.com')
-    assert "urlopen error" in out
+    assert "[Errno 8] nodename nor servname provided, or not known" in out
```
BeyondEvil: The error message was different now due to Urllib3.

tox.ini
```diff
@@ -9,7 +9,7 @@ envlist = py{27,36,37,py,py3}, docs, flake8
 [testenv]
 passenv = DISPLAY PYTEST_ADDOPTS
 deps = pytest-localserver
-commands = pytest -v -r a {posargs}
+commands = pytest -W error::DeprecationWarning -v -r a {posargs}
```
BeyondEvil: This will cause any test to Fail if they cause any DeprecationWarnings to be invoked. @davehunt

testing/conftest.py
davehunt: Thanks @BeyondEvil. This was introduced in pytest 3.0.0, which we already depend on, so this is a very safe change.

pytest_selenium/pytest_selenium.py
```diff
@@ -123,7 +123,6 @@ def driver_kwargs(request, capabilities, chrome_options, driver_args,
         host=pytestconfig.getoption('host'),
         port=pytestconfig.getoption('port'),
         request=request,
-        log_path=None,
```
davehunt: I think we should change this to `service_log_path` instead of removing it. Otherwise a `driver_log` of `None` would result in the default Firefox driver's log path of `geckodriver.log` to be used.

BeyondEvil:
I kept the `driver_log` fixture as to not introduce a breaking change for anyone that has overloaded it or uses directly.

pytest_selenium/drivers/edge.py
```diff
@@ -1,14 +1,20 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+from distutils.version import LooseVersion
+from selenium import __version__ as SELENIUM_VERSION
 
 
-def driver_kwargs(capabilities, driver_log, driver_path, **kwargs):
-    kwargs = {}
+def driver_kwargs(capabilities, driver_path, service_log_path, **kwargs):
```
davehunt: I'd like to keep the keyword argument as `driver_log` as this is already our abstraction from `log_path` and `service_log_path`.

pytest_selenium/pytest_selenium.py
```diff
@@ -116,14 +116,13 @@ def driver_kwargs(request, capabilities, chrome_options, driver_args,
         capabilities=capabilities,
         chrome_options=chrome_options,
         driver_args=driver_args,
-        driver_log=driver_log,
         driver_path=driver_path,
         firefox_options=firefox_options,
         firefox_profile=firefox_profile,
         host=pytestconfig.getoption('host'),
         port=pytestconfig.getoption('port'),
+        service_log_path=driver_log,
```
davehunt: The idea here is that `driver_log` gets used within our `driver_kwargs` methods to update `service_log_path`, which should default to `None`.

testing/conftest.py
davehunt: I'm going to be a pain and ask you to split this (and a few other things) out into a separate commit or pull request. This shouldn't break anything for anyone, but if it does it makes it easier to identify if it's not combined with an unrelated change.

testing/test_crossbrowsertesting.py
davehunt: Could you move these test fixes into a separate commit or pull request?

testing/test_safari.py
```diff
@@ -9,7 +9,7 @@
 
 @pytest.mark.safari
 def test_launch(testdir, httpserver):
-    httpserver.serve_content(content='<h1>Success!</h1>')
+    httpserver.serve_content(content='<body><h1>Success!</h1></body>')
```
davehunt: Could you make this a separate commit or pull?

tox.ini
```diff
@@ -8,6 +8,7 @@ envlist = py{27,36,37,py,py3}, docs, flake8
 
 [testenv]
 passenv = DISPLAY PYTEST_ADDOPTS
+setenv = MOZ_HEADLESS=1
```
davehunt: Could you make this a separate commit or pull?

PR #183, opened 2018-08-22
Update deprecated functions and handle internal deprecations

BeyondEvil:


pytest_selenium/pytest_selenium.py
```diff
@@ -93,13 +97,27 @@ def capabilities(request, driver_class, chrome_options, firefox_options,
...
     return capabilities
 
 
+def _get_marker(node, name):
+    # get_marker is deprecated
+    # https://docs.pytest.org/en/latest/mark.html#marker-revamp-and-iteration
+    try:
+        markers = dict()
+        for each in node.iter_markers(name):
+            markers.update(each.kwargs)
```
davehunt: As I understand it, this will find the `capabilities` markers applied to the various scopes and methods in use. The capabilities from each marker will be updated by the next. Whilst this it probably what we should have been doing all along, it could cause some confusion and potential breakage to existing users. I would suggest adding debug level logging for each marker found with as much context as we can. That way, a user can turn on debug logging and understand where/why any capabilities are being overwritten.

pytest_selenium/pytest_selenium.py
```diff
@@ -93,13 +97,27 @@ def capabilities(request, driver_class, chrome_options, firefox_options,
...
+        return markers
+    except AttributeError:  # backwards-compat
+        marker = node.get_marker(name)
```
davehunt: Could you mention when we could remove this? Once we require a certain minimum version of pytest, we can simplify this code.

pytest_selenium/pytest_selenium.py
```diff
@@ -93,13 +97,27 @@ def capabilities(request, driver_class, chrome_options, firefox_options,
...
+    capabilities_marker = _get_marker(request.node, 'capabilities')
+    LOGGER.debug('Capabilities from marker: {}'.format(capabilities_marker))
+    if capabilities_marker:
         # add capabilities from the marker
-        capabilities.update(capabilities_marker.kwargs)
+        capabilities.update(capabilities_marker)
     return capabilities
 
 
+def _get_marker(node, name):
```
davehunt: This isn't returning the markers, but a dictionary of the marker's keyword arguments. I appreciate your desire to make this a generic function, but I think in this case we could just name it `get_capabilities_from_markers`. We could also simplify to `capabilities.update(get_capabilities_from_markers(request.node))` as the return value will always be a dictionary so we don't need the `if x is not None` block.

testing/conftest.py
```diff
@@ -34,6 +34,14 @@ def webtext(base_url, selenium):
...
         """)
 
+    testdir.makefile('.cfg', setup="""
+        [tool:pytest]
+        filterwarnings =
+            error::DeprecationWarning
+            ignore:--firefox-\w+ has been deprecated:DeprecationWarning
+            ignore:MarkInfo:DeprecationWarning:pytest_selenium.drivers.firefox:88
+    """)
+
```
davehunt: Nice! What do we need to do to enable failure on warnings?

pytest_selenium/pytest_selenium.py
BeyondEvil: I did some testing with module and function level markers, and that code behaves just as `get_markers()` did. That is, if there's a conflict between module and function, the function wins.

With that said, I can see if I can add more debugging.

testing/conftest.py
BeyondEvil: Do you want it to fail on all warnings, not just DeprecationWarnings?

davehunt: Let's discuss in a new issue/pull request. I'm thinking just deprecation warnings, but also interested to know what other warnings we have.

pytest_selenium/pytest_selenium.py
davehunt: Oh so we don't iterate through them? That's interesting.. then what are we iterating through? Multiple markers with the same name on a single object?

PR #212, opened 2019-01-21
Fix `is not 'literal'`

asottile:
This will be a new lint error in the latest pyflakes / flake8

### DIFF
--- a/pytest_selenium/drivers/crossbrowsertesting.py
+++ b/pytest_selenium/drivers/crossbrowsertesting.py
@@ -78,7 +78,7 @@ def pytest_selenium_runtest_makereport(item, report, summary, extra):
 
     try:
         # Update the test result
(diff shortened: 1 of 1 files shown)


BeyondEvil:
@RonnyPfannschmidt @davehunt @nicoddemus 

Do any of you guys have any idea why the xvfb fails to start on Travis?

`The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .`

BeyondEvil:
Found the reason:

`Xvfb init file is removed in favor of a systemd unit file. Xvfb can be set up with services: xvfb. The matching DISPLAY environment variable is made available when you do that.`

https://docs.travis-ci.com/user/build-environment-updates/2019-01-14/

Will create a PR with a fix. @asottile

Issue #216, opened 2019-02-06
TestingBot tunnel tests fails with SSL error


BeyondEvil:
When trying to run tests against a local webserver using `--host localhost` and `--port 4445` as per the TestingBot [docs](https://testingbot.com/support/other/tunnel), the tests errors out with: 

`urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='localhost', port=4445): Max retries exceeded with url: /wd/hub/session (Caused by SSLError(SSLError(1, '[SSL: UNKNOWN_PROTOCOL] unknown protocol (_ssl.c:1045)')))`

To make this work, the `executor` url needs to be "http".

PR with fix is incoming.

PR #217, opened 2019-02-06
Fix issue with using localhost for TestingBot tunnel

BeyondEvil:
Fixes: #216

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -5,6 +5,7 @@ Release Notes
 -------------------
 
 * ``pytest-selenium`` now requires pytest 3.6 or later.
+* Fixed `issue <https://github.com/pytest-dev/pytest-selenium/issues/216>`_ with TestingBot local tunnel.
 
 1.15.1 (2019-01-07)
 -------------------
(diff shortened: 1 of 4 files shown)


Dude-X:
Is it possible that the port is 443 the standard HTTPS port and no longer custom? I would prefer to try to maintain encryption.

BeyondEvil:
> Is it possible that the port is 443 the standard HTTPS port and no longer custom? I would prefer to try to maintain encryption.

When using TestingBot without a tunnel we use HTTPS and port 443. But what the tunnel does is it enables you to let the TestingBot VM access any webserver running locally on your machine (and not exposed to the public internet).

The [tunnel](https://testingbot.com/support/other/tunnel) itself is secure and uses SSH to communicate with the TestingBot VM. But for the tests to be able to communicate with the VM (via the tunnel) you have to point the tests to `localhost` and port `4445` (default, configurable).

So the "unsecured" http-traffic is happening only on your local machine, never with the TB VM.

Testing bot [example](https://testingbot.com/support/other/tunnel/examples.html)

Our [docs](https://pytest-selenium.readthedocs.io/en/latest/user_guide.html#local-tunnel)

Let me know if that clears things up @Dude-X

Dude-X:
Thanks for the clarification.

BeyondEvil:
Sorry, had to add the changes to `news.rst` because I want to release a new version asap, to fix this bug. @Dude-X

PR #222, opened 2019-06-04
Getting appveyor to work take 1

BeyondEvil:
Get appveyor working for IE and maybe Edge tests.


BeyondEvil:
Since we have no tests for IE and Appveyor doesn't support Edge, we drop it.

PR #223, opened 2019-06-13
Deprecate phantom

BeyondEvil:
There is no need for PhantomJS anymore since both Chrome and Firefox support headless testing. I suggest we deprecate it and remove it in the next major release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,8 @@
 language: python
 jobs:
   include:
-    - stage: Linting
+    - stage: Tests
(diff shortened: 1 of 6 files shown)


BeyondEvil:
Changes to `.travis.yml` are not related to deprecating PhantomJS, but makes for a [prettier](https://travis-ci.org/pytest-dev/pytest-selenium/builds/545138294?utm_source=github_status&utm_medium=notification) layout. 😁

Issue #133, opened 2017-09-13
Save screenshot to tempfile in text mode


xmo-odoo:
Screenshots are really valuable tools for debugging selenium test (even more so when running in headless Firefox or Chrome). Currently, pytest-selenium injects screenshots into HTML reports, but maybe it could also save those screenshots to the disk and give the file path in the text summary, possibly optionally? At the moment it generates a (base64) screenshot in every case and just drops it when `html` is not enabled.

ramonasuciu:
I was able to save the screenshots to disk, using Selenium's save_screenshot(file_path).
You can rename the files as you wish and move them to a separate folder. I left them in assets, for convenience purposes, but the idea is that, using save_screenshot(file_path), you are able to keep the screenshots locally. Also, you can inject them still in pytest's html report. 

Hope it helps.

xmo-odoo:
@ramonasuciu yeah you can take them manually without issues, but it's less convenient than pytest-selenium doing so on only failures, which as far as I can see it already does *but* it drops the screenshot if not using HTML reports.

ramonasuciu:
@xmo-odoo what I meant is that pytest-html still picks the screenshots which are now created as png files, not as base64, and attaches them to the corresponding failing scenario in the html report.
The only difference is that I am now saving the screenshots as files per-say, in any location, and instruct pytest to pick them up from that new location, in order to properly attach them to the html report.

Please let me know if you'd need more details on how I've done this.

Thanks,
Ramona

BeyondEvil:
@davehunt What if were to, in the case of pytest-html not being enabled, drop the files to disk and put the file path in the summary? Either per default or per the command line/ini-setting.

ping @xmo-odoo

xmo-odoo:
@BeyondEvil that's pretty much my original suggestion.

BeyondEvil:
@xmo-odoo Yes :)

The problem I can see with dumping the file being the default, is littering the filesystem with screenshots.

davehunt:
[...]

PR #226, opened 2019-07-01
Add tips and tricks section to user guide

BeyondEvil:
What we talked about @Dude-X 

Sometimes it's best to have suggested solutions to common scenarios, instead of trying to shoehorn them into to the plugin.

Fixes: #133

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -48,7 +48,7 @@ master_doc = "index"
 # General information about the project.
 project = u"pytest-selenium"
 copyright = u"2019, Dave Hunt"
-author = u"Dave Hunt"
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -47,7 +47,7 @@
 
 # General information about the project.
 project = u"pytest-selenium"
-copyright = u"2015, Dave Hunt"
+copyright = u"2019, Dave Hunt"
 author = u"Dave Hunt"
```
davehunt: I think we should add other authors here now.. my last commit was almost a year ago.

Dude-X:
looks good to me. I would probably update this in the future once python 3 is the norm. I think the encode command is for py2 compatibility.

PR #228, opened 2019-09-16
Avoid importing requests for test runs that don't need it

boxed:
The import of requests is pretty expensive, and if pytest-selenium is just installed as a plugin we incur the overhead for that, even though we're running unit tests that don't need it. 

The moved imports and copied CaseInsensitiveDict is a bit uglier but saves 0.1-0.2 seconds on my machine.

### DIFF
--- a/pytest_selenium/drivers/browserstack.py
+++ b/pytest_selenium/drivers/browserstack.py
@@ -3,7 +3,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import pytest
-import requests
 
 from pytest_selenium.drivers.cloud import Provider
 
@@ -42,6 +41,10 @@ def pytest_selenium_runtest_makereport(item, report, summary, extra):
     passed = report.passed or (report.failed and hasattr(report, "wasxfail"))
     session_id = item._driver.session_id
     api_url = provider.API.format(session=session_id)
+
+    # import requests lazily here to avoid importing
+    # requests for unit tests that don't need it
+    import requests
 
     try:
         job_info = requests.get(api_url, auth=provider.auth, timeout=10).json()
--- a/pytest_selenium/drivers/crossbrowsertesting.py
+++ b/pytest_selenium/drivers/crossbrowsertesting.py
@@ -4,7 +4,6 @@
 
 from py.xml import html
 import pytest
-import requests
 
 from pytest_selenium.drivers.cloud import Provider
 
@@ -40,6 +39,10 @@ def pytest_selenium_capture_debug(item, report, extra):
     if not provider.uses_driver(item.config.getoption("driver")):
         return
 
+    # import requests lazily here to avoid importing
+    # requests for unit tests that don't need it
+    import requests
+
     videos = (
         requests.get(
             provider.API.format(session=item._driver.session_id),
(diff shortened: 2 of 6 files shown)


boxed:
A further optimization here might be to not import selenium. But this is a bigger impact change that I wasn't comfortable doing at this point. But just importing selenium is ~0.2s so it would be nice to not do that just because this plugin is installed.

Dude-X:
I am a -1 on this change. It creates a dependency on updating the dictionary if there's a change for whatever reason, and it's a micro-optimization that adds more code to save a few tenths of a second. 

Other core people might feel different.

Dude-X:
Also in Python, an import occurs once for a process, so once it's imported, there is no penalty for repeated imports.

boxed:
I'm confused. Your first comment sounds like it talks about the suggested future change, but it says "this change". Can you clarify?

And yes I know how imports work. The thing with tests though is that often one runs just one test. And the way this plug-in is written and the way pytest works you have to have the plug-in installed always if you need it for just a few tests and you always incur the cost even when running tests that don't use it. 

It's a micro optimization sure. But this is a plug-in that is used in lots of projects so it's a micro win multiplied by a huge number. 

Also I maintain mutmut, a mutation tester, and there the process overhead is paid for every mutant. That adds up! (I'm trying to combat this)

Dude-X:
Based on this feedback, I am a +0 on your PR. You make a good point that all plugins are loaded when running pytest in a particular environment, so that does always create higher overhead for a test suite.

I am curious to see what benches are like in mutmut with and without the patch.

boxed:
Well, the problem is I have to land similar patches in several other plug-ins and it's still terrible performance for my work project :) But it's a start!

Dude-X:
It may be ambitious to have every project you use resort to lazy loading. It may be better to use custom pytest.ini files where you specify `addopts = -p no:pytest-selenium` etc to disable plugins you don't want loaded.

Alternatively see if you can make pluggy better.

boxed:
It's not that bad. I've sent prs already and the upside is that these are all super common plugins so we can fix this for the entire python ecosystem. That would be great!

BeyondEvil:
@boxed 

Did you make any headway with the other PRs? Is this still a pain point for you?

boxed:
I have! And it is.

BeyondEvil:
Ok, cool! :)  @boxed 

If I get buy in from other maintainers (see #233), I might make some changes that should speed things up.

Dude-X:
In light of the fact that requests is no longer a hard requirement for the main `selenium` library, then it should work for users who don't depend on the 3rd party grids. So I am now +1 on this change.

BeyondEvil:
Hey @boxed 

I'm going to merge this and release it in the next release. However, I want a couple of things to happen first.
1. You need to rebase with master.
2. I don't want the `CaseInsensitiveDict` class in the main plugin file (`pytest_selenium.py`). It's not relevant to the logic of the plugin and just adds cognitive overhead. Move it and import it, please.

Thanks!

boxed:
Huh.. I don't know how actually. I seem to have removed my github fork of pytest-selenium, I don't have it on my machine anymore either. Oops! 

I can remake the changes like you want on master and send a new PR I guess. Should I do that?

Dude-X:
I can take care of merging and making the appropriate changes if it’s too
much trouble to redo.

boxed:
That would be great! The changes aren't really big.

Dude-X:
@boxed can you give us permission to modify your PR? otherwise I would have to redo your work in my own fork.

boxed:
Ehm, how? My fork is gone so I think that's it.

Dude-X:
I am going to redo your code in my fork so that it just works. But I just need your approval to go ahead and do this.

boxed:
You have it.

BeyondEvil:
Closing this in favor of #246

PR #240, opened 2020-06-15
Fix report tests for pytest-html >= 2.0.0

PhilippSelenium:


BeyondEvil:
Thanks for this!

But I consider the cause of this a bug in pytest-html. So I've created a new PR to pin the version. Unfortunately Github is experiencing some issues, so I can't link to it.

Please let me know if you disagree.

testing/test_report.py
```diff
@@ -5,16 +5,24 @@
...
 
 pytestmark = pytest.mark.nondestructive
 
 URL_LINK = '<a class="url" href="{0}/" target="_blank">URL</a>'
-SCREENSHOT_LINK_REGEX = (
-    '<a class="image" href=".*" target="_blank">Screenshot</a>'
-)  # noqa
-SCREENSHOT_REGEX = '<div class="image"><a href=".*"><img src=".*"/></a></div>'
+
+if LooseVersion(PYTEST_HTML_VERSION) < LooseVersion("2.0.0"):
```
BeyondEvil, line 17: It's actually less than `v2.1.0`, but I don't think that matters right now.

PR #243, opened 2020-06-19
Limit pytest-html version due to bug in >2.0.1

BeyondEvil:
There was a [bug](https://github.com/pytest-dev/pytest-html/issues/308) introduced in pytest-html v2.1.0, so we won't use that (or newer) until it has been fixed.

Would people prefer this or #240 ? 🤔

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -334,7 +334,11 @@ def _gather_logs(item, report, driver, summary, extra):
 
 def _gather_driver_log(item, summary, extra):
     pytest_html = item.config.pluginmanager.getplugin("html")
-    if hasattr(item.config, "_driver_log") and os.path.exists(item.config._driver_log):
+    if (
+        hasattr(item.config, "_driver_log")
+        and item.config._driver_log is not None
+        and os.path.exists(item.config._driver_log)
+    ):
         if pytest_html is not None:
             with io.open(item.config._driver_log, "r", encoding="utf8") as f:
                 extra.append(pytest_html.extras.text(f.read(), "Driver Log"))
@@ -344,7 +348,7 @@ def _gather_driver_log(item, summary, extra):
 def format_log(log):
     timestamp_format = "%Y-%m-%d %H:%M:%S.%f"
     entries = [
-        u"{0} {1[level]} - {1[message]}".format(
+        "{0} {1[level]} - {1[message]}".format(
             datetime.utcfromtimestamp(entry["timestamp"] / 1000.0).strftime(
                 timestamp_format
             ),
--- a/testing/test_crossbrowsertesting.py
+++ b/testing/test_crossbrowsertesting.py
@@ -56,28 +56,3 @@ def test_missing_access_key_file(failure, monkeypatch, tmpdir):
     monkeypatch.setattr(os.path, "expanduser", lambda p: str(tmpdir))
     tmpdir.join(".crossbrowsertesting").write("[credentials]\nusername=foo")
     assert "CrossBrowserTesting key must be set" in failure()
-
-
-@pytest.mark.parametrize(
-    ("username", "key"),
-    [
-        ("CROSSBROWSERTESTING_USERNAME", "CROSSBROWSERTESTING_AUTH_KEY"),
-        ("CROSSBROWSERTESTING_USR", "CROSSBROWSERTESTING_PSW"),
-    ],
-)
-def test_invalid_credentials_env(failure, monkeypatch, tmpdir, username, key):
-    monkeypatch.setattr(os.path, "expanduser", lambda p: str(tmpdir))
-    monkeypatch.setenv(username, "foo")
-    monkeypatch.setenv(key, "bar")
-    out = failure("--capability", "browser_api_name", "FF46")
-    messages = ["missing auth", "basic auth failed"]
-    assert any(message in out for message in messages)
-
-
-def test_invalid_credentials_file(failure, monkeypatch, tmpdir):
-    monkeypatch.setattr(os.path, "expanduser", lambda p: str(tmpdir))
-    config = tmpdir.join(".crossbrowsertesting")
-    config.write("[credentials]\nusername=foo\nkey=bar")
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -14,7 +14,7 @@ setenv =
 deps =
     pytest-localserver
     pytest-xdist
-commands = pytest -n auto -v -r a {posargs}
+commands = pytest -n auto -v -r a --color=yes --html={envlogdir}/report.html --self-contained-html {posargs}
```
BeyondEvil, line 17: Useful when running tests locally (for now).

PhilippSelenium:
@BeyondEvil Could you reference the mentioned pytest-html bug?

BeyondEvil:
> @BeyondEvil Could you reference the mentioned pytest-html bug?

I've updated the description. But for reference: https://github.com/pytest-dev/pytest-html/issues/308

testing/test_crossbrowsertesting.py
```diff
@@ -71,7 +71,7 @@ def test_invalid_credentials_env(failure, monkeypatch, tmpdir, username, key):
     monkeypatch.setenv(key, "bar")
     out = failure("--capability", "browser_api_name", "FF46")
     messages = ["missing auth", "basic auth failed"]
-    assert any(message in out for message in messages)
+    assert any(message in out.lower() for message in messages)
```
BeyondEvil, line 74: What are we even testing here? The error message from CrossBrowserTesting when creds are bad. Why? The plugin doesn't handle that anyway. 🤔

nicoddemus:
Limiting pytest-html version or #240 are both fine I guess, given the current version is already broken when used with current pytest-html.

BeyondEvil:
Ok, here's my thinking. @nicoddemus 

I like @PhilippSeleniums solution because it doesn't silently limit the version of `pytest-html` that a user would get when installing `pytest-selenium`. I don't think that the bug is significant enough that a user normally would care enough to want to limit the version because of it.

So with that in mind, what I can do is update this PR to issue a warning instead that they're using a version with a known bug (and link to the issue here in GH). They can then pin the version themselves if it in fact does impact them.

Does that make sense?

testing/test_crossbrowsertesting.py
```diff
@@ -80,4 +80,4 @@ def test_invalid_credentials_file(failure, monkeypatch, tmpdir):
     config.write("[credentials]\nusername=foo\nkey=bar")
     out = failure("--capability", "browser_api_name", "FF46")
     messages = ["missing auth", "basic auth failed"]
-    assert any(message in out for message in messages)
+    assert any(message in out.lower() for message in messages)
```
BeyondEvil, line 83: @nicoddemus I'll remove these tests instead. They're basically testing that CBT returns an error with incorrect credentials.

A case we don't handle explicit anyway. So I don't really see the point of them. 🤔 

I even think it was an earlier version of me that implemented them (or at least some of them).

BeyondEvil:
Thanks @nicoddemus, you’re so supportive! <3

BeyondEvil:
Done @nicoddemus 

Here's what the output is under the warnings summary:

```
test_html.py::test_url
  /Users/jimbrannlund/dev/pytest-dev/pytest-selenium/pytest_selenium/pytest_selenium.py:244: UserWarning: pytest-html version 2.1.0 contains a bug that causes the screenshot href to not be named. See: https://github.com/pytest-dev/pytest-html/issues/308
    f"pytest-html version {PYTEST_HTML_VERSION} contains "

-- Docs: https://docs.pytest.org/en/latest/warnings.html
```

Why is this line printed out: `f"pytest-html version {PYTEST_HTML_VERSION} contains "`

`warnings` seems to always print do that. A bug?

pytest_selenium/pytest_selenium.py
```diff
@@ -344,7 +353,7 @@ def _gather_driver_log(item, summary, extra):
 def format_log(log):
     timestamp_format = "%Y-%m-%d %H:%M:%S.%f"
     entries = [
-        u"{0} {1[level]} - {1[message]}".format(
+        "{0} {1[level]} - {1[message]}".format(
```
BeyondEvil, line 356: formatting done by `black`

nicoddemus:
> Why is this line printed out: f"pytest-html version {PYTEST_HTML_VERSION} contains "

That's the line of code on the warnings line, I would guess. I wonder though why it is not showing up the message itself...

BeyondEvil:
> > Why is this line printed out: f"pytest-html version {PYTEST_HTML_VERSION} contains "
> 
> That's the line of code on the warnings line, I would guess. I wonder though why it is not showing up the message itself...

It did tho:
```
UserWarning: pytest-html version 2.1.0 contains
```

BeyondEvil:
If everything looks good, can I trouble you for a approval @nicoddemus ?

BeyondEvil:
Fucking hell... it never ends does it?

I have no idea what's happening.

pytest_selenium/pytest_selenium.py
```diff
@@ -237,6 +238,14 @@ def pytest_report_header(config, startdir):
 
 @pytest.mark.hookwrapper
 def pytest_runtest_makereport(item, call):
+    from pytest_html import __version__ as PYTEST_HTML_VERSION
+
+    if PYTEST_HTML_VERSION is not None and PYTEST_HTML_VERSION >= "2.1.0":
```
nicoddemus, line 243: Sorry for missing this earlier, you are better off using `disutils.version` for comparison here instead of comparing the strings, as doing so does not produce correct results for all versions.

Is the intention here to change this warning later to also suggest installing a `pytest-html` version with the bug fixed?

Alternatively it would probably be simpler to just fix https://github.com/pytest-dev/pytest-html/issues/308 instead and make a new release, and don't even issue this warning at all? 🤔

BeyondEvil: Yeah, you're right. This was supposed to be a quick fix because it was blocking. I got tunnelvision.

It does however contain some needed fixes, so I'll update.

BeyondEvil:
This is ready now. Thanks @nicoddemus

Issue #233, opened 2020-03-15
[RFC] Dropping python 2.7 support and other changes


BeyondEvil:
I'm thinking it's time for a 2.0 release of the plugin.

Here's what I'm thinking we can do.

1. Drop support for python 2.7
1. Remove phantomjs as a driver
1. Make all the cloud providers optional (same as `appium`)
1. (Maybe make `requests` a "soft" requirement, see [here](https://github.com/pytest-dev/pytest-selenium/pull/228) for reasoning, not implementation)

Thoughts? Anything else we can do?

Thanks!

ping @davehunt @nicoddemus @RonnyPfannschmidt

BeyondEvil:
Maybe you have some input as well @Dude-X ? 😊

nicoddemus:
Sounds good. 👍

Dude-X:
@BeyondEvil I wouldn't drop Python 2.7 support unless we are using an API that is in the later version of PyTest that doesn't support Python 2.7. Until there is an API change that forces us to update the minimum version, then we can drop Python 2.7 support.

Dude-X:
> Remove phantomjs as a driver

This seems reasonable since the project is dead.

>Make all the cloud providers optional (same as appium)

Do you mean make it installable via the extras keyword? I would probably seek feedback from the community.

>(Maybe make requests a "soft" requirement, see here for reasoning, not implementation)

This would make sense if we do separate the cloud providers. Even if we don't, we can make sure requests isn't imported if there is no need for it.

BeyondEvil:
Thanks for your feedback @Dude-X, I appreciate it! 🙏 

> @BeyondEvil I wouldn't drop Python 2.7 support unless we are using an API that is in the later version of PyTest that doesn't support Python 2.7. Until there is an API change that forces us to update the minimum version, then we can drop Python 2.7 support. We've already done that in `pytest-html`.

Right, as part of dropping python 2.7 support, I was thinking we up the minimum required pytest to a version that corresponds to that.

> > Make all the cloud providers optional (same as appium)
> 
> Do you mean make it installable via the extras keyword? I would probably seek feedback from the community.

Exactly, as it stands right now - you get all the providers installed, when it's unlikely you're using more than one. It might even be unusual that a majority of users uses even one of them - but I have no way of checking that afaik.

As for seeking feedback, any ideas of how to do that? Other than Slack?

Dude-X:
[...]

PR #247, opened 2020-07-13
Drop python 2.7 support (incl. PyPy)

BeyondEvil:
Also in this PR:
- Add support for python 3.8 (incl. CI)
- Drop support for PhantomJS

See #233 for discussion.


BeyondEvil:
I'll give it a couple days before merging to get other reviewers a chance to chime in.

Dude-X:
I think we should go ahead with this.

PR #248, opened 2020-07-17
PhantomJS cleanup

BeyondEvil:


testing/conftest.py
```diff
@@ -51,7 +51,6 @@ def chrome_options(chrome_options):
         filterwarnings =
             error::DeprecationWarning
             ignore:--firefox-\w+ has been deprecated:DeprecationWarning
-            ignore:Support for PhantomJS:DeprecationWarning
```
nicoddemus, line 54: Has this warning been removed elsewhere?

BeyondEvil: You mean in other projects? @nicoddemus 

The warning is from [here](https://github.com/pytest-dev/pytest-selenium/blob/v1.17.0/docs/user_guide.rst#phantomjs).

Issue #219, opened 2019-05-02
Please add support for retrying driver initialization


DanaSilerRhiza:
One of our build environments is having an intermittent problem where the pytest-selenium driver fixture fails due to a problem with bringing up Chrome (the error during the call to the driver fixture is a 104 - 'Connection reset by peer').

Granted, we want to get to the bottom of the intermittent initialization problem, but it would be awfully nice if we had a way to specify a number of retry attempts to make when initializing the driver. This would prevent intermittent test failures due to a problem that is not related to the code under test. 

I have worked around the lack of support for this behavior by copy/pasting the driver fixture code here: https://github.com/pytest-dev/pytest-selenium/blob/master/pytest_selenium/pytest_selenium.py#L180 into my own method. Then, I created a new fixture which has the retry logic that I want and which calls the copy/pasted pytest_selenium driver code. 

I anticipate that this solution would work for our team for a while, but I have concerns that it might break when pytest-selenium is updated. Also, other users of the pytest-selenium community might find this retry functionality useful. 

Here is my workaround for the lack of a retry at the pytest-selenium fixture level:

```   
     # Initialize the webdriver and catch any errors.
     #
     # This code came from the pytest_selenium module's driver method.
     # This method is responsible for initializing the driver. I would prefer
     # to call this method directly, but since the method is a fixture, I cannot
     # do so. The original code can be found here:
     # https://github.com/pytest-dev/pytest-selenium/blob/master/pytest_selenium/pytest_selenium.py#L181
     #
     # The initialization of the webdriver code and browser happens in the call to
     # driver_class(**driver_kwargs). The rest of the code was copied here verbatim
     # to ensure correct initialization of the returned driver.
     #
   
    def my_driver(self, request, driver_class, driver_kwargs):
        """Returns a WebDriver instance based on options and capabilities"""
        driver = driver_class(**driver_kwargs)
        
        event_listener = request.config.getoption("event_listener")
        if event_listener is not None:
            # Import the specified event listener and wrap the driver instance
            mod_name, class_name = event_listener.rsplit(".", 1)
            mod = __import__(mod_name, fromlist=[class_name])
            event_listener = getattr(mod, class_name)
            if not isinstance(driver, EventFiringWebDriver):
                driver = EventFiringWebDriver(driver, event_listener())

        request.node._driver = driver
        return driver

    @pytest.fixture(autouse=True, scope='function')
    def init_driver(self, request, driver_class, driver_kwargs):
        """
[...]

PR #249, opened 2020-07-20
Add retrying of failed driver initialization

BeyondEvil:
Fixes: #219

### DIFF
--- a/testing/test_driver.py
+++ b/testing/test_driver.py
@@ -4,7 +4,10 @@
 
 from functools import partial
 
+import pytest_selenium
 import pytest
+
+from pytest_selenium.utils import CaseInsensitiveDict
 
 pytestmark = pytest.mark.nondestructive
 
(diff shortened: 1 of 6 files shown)


testing/test_driver.py
```diff
@@ -180,3 +181,70 @@ def test_pass(driver_kwargs):
...
+
+        class MockDriver():
+            def __init__(*args, **kwargs):
+                pass
+            def quit(*args, **kwargs):
+                pass
+
+        @pytest.fixture(scope="session")
+        def driver_class():
+            return MockDriver
```
BeyondEvil, line 200: I tried using the mocker to avoid having these, but couldn't figure out how.

I tried:

```python
mocker.patch("pytest_selenium.pytest_selenium.webdriver")
```

and `"pytest_selenium.pytest_selenium.webdriver.Firefox"`

but couldn't get it to work 🤷‍♂️

Dude-X: I don't have too much experience with using `mock` but I think a better test is to use `Remote` webdriver, and then  mock the `RemoteConnection` class to throw an exception. This would be more realistic than having Firefox fail. Either way, focus on remote connection failing.

BeyondEvil: That's genius! 

But how do I accomplish that?

I tried
```python
    mocker.patch(
        "pytest_selenium.pytest_selenium.webdriver.remote.remote_connection.RemoteConnection",
        side_effect=Exception("Connection Error"),
    )
```

But it doesn't work. With that I mean, I don't get any syntax errors - but the test doesn't produce the expected outcome (the exception isn't raised.

BeyondEvil: I'm struggling to get something to work.

I managed to mock the `execute` method on `RemoteConnection`, but that slowed down the tests (from ~1s to ~6s).

Maybe @nicoddemus can share some wisdom?

BeyondEvil: Think I solved it. 👍

nicoddemus: Sorry I didn't get here sooner.

Was the fix moving construction of `driver_class` inside the `with` statement?

BeyondEvil: No worries @nicoddemus Sorry I squashed the commits and hid the history.

But no, the fix was doing the patching in the proper places.

BeyondEvil:
ping @Dude-X @nicoddemus

testing/test_driver.py
```diff
@@ -180,3 +183,60 @@ def test_pass(driver_kwargs):
...
+        import pytest
+
+        @pytest.mark.nondestructive
+        def test_pass(driver):
+            assert True  # catch if this starts to pass
+        """
+    )
+
+    testdir.quick_qa("--driver", "Firefox", file_test, failed=1)
+    assert (
```
nicoddemus, line 238: Minor: this was a bit hard to parse because it ended up being formatted by black, perhaps it could be extracted to a local variable:

```python
expected_attempts = max_num_attempts if max_num_attempts != "default" else 3
assert mock_retrying.spy_return.statistics["attempt_number"] == expected_attempts
```

BeyondEvil: Good idea!

I refined it a little bit, hope you don't mind. 😊 @nicoddemus

Issue #215, opened 2019-02-01
argparse.ArgumentError: argument --host: conflicting option string: --host


POD666:
Hi, 
here is the issue I've faced:
```
$ python3 -V
Python 3.7.0
$ python3 -m venv testenv
$ source testenv/bin/activate 
$ pip install pytest-selenium
...
Successfully installed atomicwrites-1.3.0 attrs-18.2.0 certifi-2018.11.29 chardet-3.0.4 idna-2.8 more-itertools-5.0.0 pluggy-0.8.1 py-1.7.0 pytest-4.2.0 pytest-base-url-1.4.1 pytest-html-1.20.0 pytest-metadata-1.8.0 pytest-selenium-1.15.1 pytest-variables-1.7.1 requests-2.21.0 selenium-3.141.0 six-1.12.0 urllib3-1.24.1
$ pytest
Traceback (most recent call last):
...
File ".../argparse.py", line 1529, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument --host: conflicting option string: --host
```
Any suggestions?

BeyondEvil:
Any news on this @POD666 ?

MidCheck:
Hello, I have encountered the same error:
```
>>> import argparse
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('-h', '--host')
Traceback (most recent call last):
...
File "/usr/lib64/python3.7/argparse.py", line 1529, in _handle_conflict_error
    raise ArgumentError(action, message % conflict_string)
argparse.ArgumentError: argument -h/--host: conflicting option string: -h
```
Finally, I think the cause of the error might be that the function add_argument() specifies the '--host' parameter, which may conflict with the default '-h' (help) parameter, see: https://docs.python.org/zh -cn/3.7/library/argparse.html

BeyondEvil:
@MidCheck 

Can you please share information regarding your environment?

OS
python version
pytest and plugin versions
anything else that you might think is pertinent to reproduce the issue.

Also, a minimal test case that reproduces the issue would help.

Thanks!

BeyondEvil:
I'm going to close this issue due to lack of interaction. If you're still seeing this and can successfully reproduce it - feel free to reopen.

topher-brennan:
[...]

PR #253, opened 2020-08-20
Rename host and port arguments for GRID

BeyondEvil:
Rename the Selenium GRID `--host` and `--port` to `--grid-host` and `--grid-port` respectively, to be more verbose and avoid collisions with other plugins.

See comment [here](https://github.com/pytest-dev/pytest-selenium/issues/215#issuecomment-676907505).

### DIFF
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -248,11 +248,11 @@ driver, as the accepted capabilities may differ::
   pytest --driver Remote --capability browserName firefox
 
 Note that if your server is not running locally or is running on an alternate
-port you will need to specify the ``--host`` and ``--port`` command line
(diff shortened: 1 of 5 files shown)


BeyondEvil:
> Before we merge this, I suggested `grid_host` but not everyone connects to a "grid" sometimes it can just be a standalone server that is on another host. so maybe `selenium_host` could be better or `webdriver_host` so this way people who run compute clusters can use `grid`.
> Thoughts?

Yes, the naming irked me, but I wasn't sure what was better. I like `webdriver`, but since the plugin name (and the env vars) use `selenium`, I'll go with that. 👍

BeyondEvil:
Done @Dude-X

PR #260, opened 2021-01-12
Browserstack little improvements and Edge fix about options

nfk:
A mixed MR for a better integration with BrowserStack.

Browserstack report is available from 2 URLs private or public. A static variable is added with a default value to keep backward compatibility with `browser_url`. Now users can override the URL field with `public_url`.

More details in commit messages.

### DIFF
--- a/pytest_selenium/pytest_selenium.py
+++ b/pytest_selenium/pytest_selenium.py
@@ -112,7 +112,7 @@ def capabilities(
             key = firefox_options.KEY
             options = firefox_options.to_capabilities()
         elif browser == "EDGE":
-            key = edge_options.KEY
+            key = getattr(edge_options, "KEY", None)
             options = edge_options.to_capabilities()
         if all([key, options]):
             capabilities[key] = _merge(capabilities.get(key, {}), options.get(key, {}))
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -378,6 +378,24 @@ See the
 for additional configuration that can be set using ``--capability`` command line
 arguments.
 
+Test result links
+~~~~~~~~~~~~~~~~~
+
+By default, links to BrowserStack jobs are only visible to users logged in to the account
+that ran the job. You can choose to display a public URL in the pytest summary.
+
+This can be configured by setting the ``BROWSERSTACK_JOB_ACCESS`` environment variable or by
+adding ``report`` section in ``.browserstack`` configuration file.
+Allowed values are ``browser_url`` for private access and ``public_url`` for everyone.
+
+An example using a configuration file:
+
+.. code-block:: ini
+
+  [report]
+  job_access = browser_url
+
+
 TestingBot
 ----------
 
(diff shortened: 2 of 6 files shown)


pytest_selenium/pytest_selenium.py
```diff
@@ -112,7 +112,6 @@ def capabilities(
             key = firefox_options.KEY
             options = firefox_options.to_capabilities()
         elif browser == "EDGE":
-            key = edge_options.KEY
```
PhilippSelenium, line 115: @nfk This was written with selenium 4.0 in mind: https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/edge/options.py

Does this brake something for you with the current selenium version?

If not we should revert this change. Else we should add a workaround checking the version or adding a try except.

nfk: As notice in commit https://github.com/pytest-dev/pytest-selenium/pull/260/commits/1443b9c0029dd06755fa9d9aa27a03c47a3a2c8d. It does not work on selenium 3.141. I am going to do the change to support selenium 3.X and 4.X. Thanks for the context.

nfk: @PhilippSelenium fixed, then I fixed the CI too.

PhilippSelenium: (y)

BeyondEvil (changes requested):
Thanks for this PR, sorry it took me some time to review it!

Could you add an ini option (maybe `browserstack_browser_url`) to set the value to either `public_url` or `private_url`(default) and also update the docs?

Thanks again!

nfk:
@BeyondEvil the browserstack url is now configurable via ini or env variable. I updated the cloud module to be able to add a new section. Let me know if the naming of the new field is good.

BeyondEvil:
> @BeyondEvil the browserstack url is now configurable via ini or env variable. I updated the cloud module to be able to add a new section. Let me know if the naming of the new field is good.

Excellent, thanks!

Naming is fine, but out of curiosity, why `browser_url` instead of `private_url`? Is it in accordance with BrowserStack documentation or something like that?

Could I also ask you to fix the broken `ref`: https://github.com/pytest-dev/pytest-selenium/blob/87bff99ddd3ce353e7745d1621de384747622b84/docs/user_guide.rst#sauce-labs ?

nfk:
> > @BeyondEvil the browserstack url is now configurable via ini or env variable. I updated the cloud module to be able to add a new section. Let me know if the naming of the new field is good.
> 
> Excellent, thanks!
> 
> Naming is fine, but out of curiosity, why `browser_url` instead of `private_url`? Is it in accordance with BrowserStack documentation or something like that?
> 
> Could I also ask you to fix the broken `ref`: https://github.com/pytest-dev/pytest-selenium/blob/87bff99ddd3ce353e7745d1621de384747622b84/docs/user_guide.rst#sauce-labs ?

According to the browserstack [API doc ](https://www.browserstack.com/docs/automate/api-reference/selenium/session#get-session-details) I kept the same properties name to be consistent.

About the broken ref on sauce-labs section, it is an issue on the github viewer (RST viewer).  It works fine on sphinx doc. Or maybe I misunderstood your request.

BeyondEvil (approved):
Great job, thanks!

PR #265, opened 2021-05-06
Selenium 4.0

dosas:
There will be quite some changes in selenium 4.0
https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES

Notes: Selenium >= 4.0.0 is only installed for python > 3.6


isaulv:
1. I don't want to drop Python 3.6 support, but on the other hand once Selenium 4.0 becomes a stable release, we may have to, but after checking trio and websockets library, it seems that they do support 3.6.1+ but David Burns mentions that the CDP library uses asyncio 3.7 specific code. It would be nice if someone can use the CDP protocol with Py3.6 and verify these claims.
2. I guess we should bump up the major version since it will break some things. Selenium 4 should be "backwards" compatible, but the next revision of the Python client will get rid of the deprecated stuff to clean up code. Simon Stewart said that Se4 should be a drop in replacement, but we need to clean up the client library in Python to make it easier to maintain and a bit more consistent. So after 4.0.0 has been patched up in broader real world use, we will introduce a version bump that will remove the deprecated code.
3. We should probably test against the last 3.x version and 4.xx for a while.
4. Hopefully this doesn't require too much work.

dosas:
@isaulv Thank you for the feedback I will look into it.

dosas:
1. Python 36 is still supported but will use a selenium version smaller than 4.0, so this should also implicitly test the 3.x version while the 4.x versions are tested using the other python versions
2. The failing cloud provider tests where the result of side effects in urllib3 when mocking ```os.path.expanduser```

dosas:
Is there any way to retrigger the tests with the RC1?

isaulv:
@dosas I apparently lack the permissions to change the CI stuff. We also need to move the automation to CircleCI since Travis is now effectively a holding asset for a corporation that doesn't care.

BeyondEvil:
I need to fix CI first before doing anything else.

I think we should release a 4.0 version of this plugin to follow the new Selenium 4.0, we can keep two parallell tracks for the time being.

Thoughts @dosas @isaulv ?

dosas:
@BeyondEvil I do not understand what you mean by two parallel tracks?

Do you need help setting up the CI, can't we use the github internal stuff or is this Circle CI?

isaulv (dismissed):

BeyondEvil:
Thanks for getting this started @dosas !

We're going to set it up so that we have selenium 3 and 4 running in parallel releases.

This way we don't have to deal with hacks and workarounds to support both in the same build.

3.0 will be a long running "legacy" branch and 4.0 will be based on master.

Hope that makes sense?

dosas:
@BeyondEvil Do you need any help on this MR?

BeyondEvil:
> @BeyondEvil Do you need any help on this MR?

Yes, it would be great if you could rebase against master and resolve the conflicts. @dosas

dosas:
@BeyondEvil I resolved the merge conflicts but as of right now the selenium version 4.0.0 is not installed:
https://github.com/pytest-dev/pytest-selenium/blob/master/setup.cfg#L38

Before I solved this by installing 4.0 for python > 3.6:
https://github.com/pytest-dev/pytest-selenium/pull/265/commits/50aee00457700bd161c88d6681851bbdc29e8628

should we add that again?

BeyondEvil:
> @BeyondEvil I resolved the merge conflicts but as of right now the selenium version 4.0.0 is not installed: https://github.com/pytest-dev/pytest-selenium/blob/master/setup.cfg#L38
> 
> Before I solved this by installing 4.0 for python > 3.6: [50aee00](https://github.com/pytest-dev/pytest-selenium/commit/50aee00457700bd161c88d6681851bbdc29e8628)
> 
> should we add that again?

Yes, please.

BeyondEvil:
Note that the reason python 3.6 doesn't fail is because it's not using Selenium 4.0. @dosas

testing/conftest.py
```diff
@@ -51,9 +52,13 @@ def chrome_options(chrome_options):
         filterwarnings =
             error::DeprecationWarning
             ignore:--firefox-\w+ has been deprecated:DeprecationWarning
+            ignore:capabilities and desired_capabilities have been deprecated, please pass in a Service object:DeprecationWarning
+            ignore:firefox_profile has been deprecated, please use an Options object:DeprecationWarning
+            ignore:Setting a profile has been deprecated. Please use the set_preference and install_addons methods:DeprecationWarning
+            ignore:Getting a profile has been deprecated.:DeprecationWarning
```
BeyondEvil, line 58: What's the effort of fixing this, so we don't get the warning to start with?

Also, we should clean out all the old deprecations (like `--firefox`), but that is probably a separate PR.

dosas: I can have a look.

BeyondEvil: Ok, thanks!

I've created a selenium-3 branch where we'll continue to support Selenium 3, while master will be the Selenium 4 branch going forward.

@dosas @isaulv Good? In that case we can merge this.

dosas: (y)

BeyondEvil (approved):
LGTM!

Thank you for this!

PR #274, opened 2021-09-09
workaround for only us-west-1

joshmgrant:
Hack to update the Sauce Labs functionality for only US-West-1 data centres.


BeyondEvil:
Is there an issue to go with this change? @joshmgrant

Issue #273, opened 2021-09-06
Allow Session Scoping?


garethmj:
Hi there.

Driven by a need to perform a large number of regression tests on a legacy app by obtaining a cookie using Selenium and then essentially calling the app's API end-points to run actual tests I wanted to be able to have the Selenium fixture last the whole test session as it is by far the slowest part of any given test.

In order to achieve that I've implemented a CLI flag to make a [dynamic scoping decision](https://docs.pytest.org/en/latest/how-to/fixtures.html#dynamic-scope) as allowed in pytest 5.2 or greater that changes the scope of the `selenium` fixture itself and all it's dependent fixtures that also need to be session scoped in order to avoid a function/session scope conflict.

I wasn't sure whether folks would feel that this is an appropriate change and, as such, didn't want to submit a PR directly to the repo without asking first.

To that end I've put up a PR on my fork of the repo to show what I've done so far: https://github.com/garethmj/pytest-selenium/pull/1

If you'd be happy, in principal, to accept a PR on this subject then I can submit a proper PR against this repo. I've run the tox tests locally and it looks good except Windows related stuff which I'm afraid I can't run at the moment.

Thanks very much in advance.

PR #275, opened 2021-09-25
Session Scope Fixture CLI Option

garethmj:
Per the explanation on this issue: https://github.com/pytest-dev/pytest-selenium/issues/273

Driven by a need to perform a large number of regression tests on a legacy app by obtaining a cookie using Selenium and then essentially calling the app's API end-points to run actual tests I wanted to be able to have the Selenium fixture last the whole test session as it is by far the slowest part of any given test.

Hopefully folks are happy with this approach. Please let me know if there might a preferred method to solve this problem.

Thanks!

### DIFF
--- a/testing/test_capabilities.py
+++ b/testing/test_capabilities.py
@@ -21,6 +21,18 @@ def testfile(testdir):
     )
 
 
+@pytest.fixture
+def testfile_scopes(testdir):
+    return testdir.makepyfile(
+        """
+        import pytest
+        @pytest.mark.nondestructive
+        def test_chrome_options(chrome_options):
+            assert chrome_options
+    """
+    )
+
+
 def test_command_line(testfile, testdir):
     testdir.quick_qa("--capability", "foo", "bar", testfile, passed=1)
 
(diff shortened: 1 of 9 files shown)


BeyondEvil:
I need to wrap my head around this. But first I need to fix CI.

BeyondEvil:
Could you please rebase? 🙏

cherubrock:
Hello. I will greatly appretiate this feature!

testing/test_capabilities.py
```diff
@@ -64,6 +76,22 @@ def capabilities():
...
+    testdir.makeconftest(
+        """
+        import pytest
+        @pytest.fixture(scope='session')
+        def chrome_options(chrome_options):
+            chrome_options.add_argument('--headless')
+            return chrome_options
+    """
+    )
+    testdir.quick_qa(testfile_scopes, failed=1)
```
dosas, line 89: @garethmj  Why do you expect this test to fail? You set the scope to session via cmd_line and the derive a fixture form chrome_options fixture which should be session scoped? 

I would expect test cases like this:

cmdline flag set (seee example): 
* session scoped fixture passes
* function scoped fixture fails
cmdline flag not set:
 * session scoped fixture fails
 * function scoped fixture passes
```
        """
        import pytest
        @pytest.mark.fixture(scope="session")
        def my_fixture(<selenium_fixture_where_scope_can_be_determined>):
            return

       # define this in testfile
       def test_stuff(my_fixture):
           assert True
    """
    testdir.quick_qa("--selenium-session-scope", passed=1)

...
        """
        import pytest
        @pytest.mark.fixture(scope="function")
        def my_fixture(<selenium_fixture_where_scope_can_be_determined>):
            return

       # define this in testfile
       def test_stuff(my_fixture):
           assert True
    """
    testdir.quick_qa("--selenium-session-scope",  failed=1)
```

Ideally <selenium_fixture_where_scope_can_be_determined> should be tested with all fixtures where the scope can be determined, so I would also recommend a new file for these scope tests.

garethmj:
Hi @BeyondEvil and @dosas.

Please accept my apologies for the delayed response to this - unfortunately I left the job where I had needed this in October so I had not really been paying attention.

I'll try to get back into this in the next week or two and have some answers to your questions and re-work as soon as I can.

Thanks for your patience and apologies again!

PR #285, opened 2022-02-23
Update browserstack.py

rohitanand1614:
proposing these changes due to new updates in the url in browserstack, also an additional argument "keep_live" should be set true is suggested by browserstack


isaulv (approved):

isaulv:
Going to merge as is. I'd imagine that this change is stable going forward as of this date.

PR #329, opened 2024-06-25
Fix Deprecation Warning for datetime.utcfromtimestamp

Mogost:


danielhav:
Hi, would it be possible to merge and release this? I would like to use pytest-selenium with py3.12 where this deprecation is in my way a bit.

